PR #46, opened 2017-06-19
Add push support

max-mironov:
Added support for callbacks for iOS and Android


max-mironov:
Closing it because `add-custom-properties` branch already contains all these changes

PR #51, opened 2017-07-21
Fix ios link when using create-react-native-app then ejecting without dash or renaming

guperrot:
create-react-native-app name-with-dashes
yarn run eject
react-native-link

Without this patch it would not find the xcodeproject folder in ios/namewithdashes

### DIFF
--- a/mobile-center-link-scripts/src/ios/MobileCenterConfig.js
+++ b/mobile-center-link-scripts/src/ios/MobileCenterConfig.js
@@ -40,7 +40,15 @@ function addConfigToProject(file) {
     return new Promise(function (resolve, reject) {
         debug(`Trying to add ${file} to XCode project`);
 
-        var globString = `**/${(pjson && pjson.name ? pjson.name : '*')}.xcodeproj/project.pbxproj`
+        var globString;
+        if (pjson && pjson.name) {
+            globString = `**/${(pjson.name.replace(/-/g, ''))}`;
+        }
+        else {
+            globString = `ios/*`;
+        }
+        globString += '.xcodeproj/project.pbxproj';
+
         var projectPaths = glob.sync(globString, { ignore: 'node_modules/**' });
 
         if (projectPaths.length !== 1) {
@@ -66,7 +74,7 @@ function addConfigToProject(file) {
                 var relativeFilePath = path.relative(path.resolve(projectPath, '../..'), file);
                 var plistPbxFile = project.addFile(relativeFilePath, project.getFirstProject().firstProject.mainGroup);
                 if (plistPbxFile === null) {
-                    debug(`Looks like ${file} was already added to ${this.projectPath}`);
+                    debug(`Looks like ${file} was already added to ${projectPath}`);
                     resolve(file);
                     return;
                 }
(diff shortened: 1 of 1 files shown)


mobile-center-link-scripts/src/ios/MobileCenterConfig.js
```diff
@@ -40,7 +40,15 @@ function addConfigToProject(file) {
...
+        var globString;
+        if (pjson && pjson.name) {
+            globString = `**/${(pjson.name.replace(/-/g, ''))}`;
+        }
+        else {
+            globString = `ios/*`;
+        }
+        globString += '.xcodeproj/project.pbxproj';
+
         var projectPaths = glob.sync(globString, { ignore: 'node_modules/**' });
```
max-mironov: Hi @guperrot, in general looks good to me. Several questions:
1) If the only reason is to get rid of dashes in name does it make sense to use the following short form:
 `var globString = **/${(pjson && pjson.name ? (pjson.name.replace(/-/g, '')) : '*')}.xcodeproj/project.pbxproj`?
Also you have replaced search pattern for `ios/*.xcodeproj` instead of usage of globstar `**/*.xcodeproj`.  
In fact for most cases it should be indeed in `ios` folder but theoretically it can be in other place. What do you think?
2) Just my interest . When tried to create app name with dashes in name using `react-native init` or `rninit init` commands but it says that `it is not valid name for a project. Please use a valid identifier name (alphanumeric). ` So the case when using dashes in names is when the app was renamed intentionally or react was added to existing app, am I correct?

panarasi: IMHO, we should look at what React-Native looks for xcode project - this should mimic that.

guperrot: @max-mironov I gave the repro in https://github.com/Microsoft/mobile-center-sdk-react-native/pull/51#issue-244533783 it's when using create-react-native app then eject.

Also the reason to use ios folder in the failover is that otherwise with ** it also matches the pods xcode project in the sub directory and thus fail.

mobile-center-link-scripts/src/ios/MobileCenterConfig.js
```diff
@@ -40,7 +40,15 @@ function addConfigToProject(file) {
     return new Promise(function (resolve, reject) {
         debug(`Trying to add ${file} to XCode project`);
 
-        var globString = `**/${(pjson && pjson.name ? pjson.name : '*')}.xcodeproj/project.pbxproj`
+        var globString;
+        if (pjson && pjson.name) {
+            globString = `**/${(pjson.name.replace(/-/g, ''))}`;
+        }
+        else {
+            globString = `ios/*`;
```
dhei: nitpick: use single quote instead of backtick when not using ES6 string interpolation.

mobile-center-link-scripts/src/ios/MobileCenterConfig.js
max-mironov: @guperrot - got it, thanks for the explanation. Good stuff and thanks for looking at this!

PR #53, opened 2017-07-22
Integrate Android and iOS SDK to 0.11.0

guperrot:
And handle the breaking changes with async APIs from that android sdk version.


guperrot:
@max-mironov the ios changes will come as a separate PR if that's ok. Can start bumping some versions in this one though.

PR #56, opened 2017-07-25
Fix early logging on ios test app

guperrot:


guperrot:
@max-mironov Yes I already asked @BretJohnson about that and I will probably make another PR for that.

PR #64, opened 2017-08-02
Fix Android Java code for react 0.47.0

guperrot:
Same code will also work on earlier versions.


mobile-center-analytics/android/src/main/java/com/microsoft/azure/mobile/react/analytics/RNAnalyticsPackage.java
```diff
@@ -40,7 +40,7 @@ public RNAnalyticsPackage(Application application, String startEnabled) {
         return modules;
     }
 
-    @Override
```
dhei: Would "No @Override to support application on react native 0.47.0 or later" be a better wording?

guperrot:
@max-mironov we depend on a minimum version of 0.19 so this callback is required when using the minimum version.

The only way to get rid of this would be to introduce a dependency to 0.47 version as a minimum which we probably don't want.

PR #68, opened 2017-08-07
[ios-native-module] Fix hasCrashedInLastSession and lastSessionCrashReport

dhei:
On Android `hasCrashedInLastSession` returns `false` or `true` And `getLastSessionCrashReport` returns null or the report object. iOS returns 0 / 1 instead of a boolean for `hasCrashedInLastSession` and returns an empty JSON object instead of null for `getLastSessionCrashReport`.

Android behavior is the correct one, this PR fixes the iOS implementation

### DIFF
--- a/mobile-center-crashes/ios/RNCrashes/RNCrashes.m
+++ b/mobile-center-crashes/ios/RNCrashes/RNCrashes.m
@@ -105,12 +105,7 @@ RCT_EXPORT_METHOD(lastSessionCrashReport:(RCTPromiseResolveBlock)resolve
 {
     void (^fetchLastSessionCrashReport)() = ^void() {
         MSErrorReport *report = [MSCrashes lastSessionCrashReport];
-        NSDictionary *jsonReport = convertReportToJS(report);
-        if (jsonReport && [jsonReport count] > 0) {
-            resolve(convertReportToJS(report));
-        } else {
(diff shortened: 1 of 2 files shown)


mobile-center-crashes/ios/RNCrashes/RNCrashes.m
```diff
@@ -105,7 +105,12 @@ - (NSDictionary *)constantsToExport
 {
     void (^fetchLastSessionCrashReport)() = ^void() {
         MSErrorReport *report = [MSCrashes lastSessionCrashReport];
-        resolve(convertReportToJS(report));
+        NSDictionary *jsonReport = convertReportToJS(report);
```
guperrot: Can convertReportToJS return nil instead? Would it make sense?

dhei: @guperrot Good point, my new commit actually fixed a corner case bug. I looked at [the other reference](https://github.com/Microsoft/mobile-center-sdk-react-native/blob/master/mobile-center-crashes/ios/RNCrashes/RNCrashesDelegate.m#L48) to `convertReportToJS` method, and found `sendAppEventWithName` method is expecting `nil` instead of an empty `NSDictionary` in the case of null parameter.

PR #69, opened 2017-08-08
Remove inclusion of RN 0.19 from build dependencies

newyankeecodeshop:
This change removes the reference to React Native 0.19 from the gradle dependencies. Including `0.19.+` causes the dependency resolver to include version 0.19.1 from maven central in addition to the version of React Native installed locally via npm. While this does not necessarily create a runtime issue, it causes extra downloads and pollutes the library list in Android Studio.
The convention in other React Native libraries, such as "react-native-code-push" is to just use the `+` alone:
https://github.com/Microsoft/react-native-code-push/blob/master/android/app/build.gradle


dhei:
@newyankeecodeshop Thanks for your PR.

PR #74, opened 2017-08-10
Prepare for v0.8.1 release

dhei:


dhei:
@guperrot updated.

guperrot:
Again for 0.11.2:

You need to rebuild RNMobileCenterShared and also commit it to use it in TestApp.

You also need to commit the Pods change in TestApp.

PR #77, opened 2017-08-11
Align and fix custom properties API with other SDKs

guperrot:


guperrot:
DemoApp can be updated only after release for custom props.

PR #84, opened 2017-08-14
update react dependencies in TestApp and DemoApp

dhei:
- update to use react-native@0.47.1 
- checkin yarn.lock
- remove yarn.lock from gitignore


dhei:
@guperrot is on vacation, @BretJohnson can you review this PR?

PR #97, opened 2017-08-30
Change jcenter mobile-center-react-native reference to exact version instead of dynamic range like "0.8.+"

dhei:


dhei:
Thanks for your feedback @jaelim-ms

PR #111, opened 2017-09-08
Add eslint to mobile-center-link-scripts

dhei:


dhei:
@guperrot I have tested `react-native link` works as expected for application with or without mobile center modules installed.

PR #120, opened 2017-09-20
Fix a typo in crashes logging

dhei:
Looks like a copy-and-paste error. `"Failed to send onBeforeSending event:"` should be `"Failed to send onSendingFailed event:"`.

### DIFF
--- a/TestApp/yarn.lock
+++ b/TestApp/yarn.lock
@@ -34,15 +34,42 @@ acorn-globals@^3.1.0:
   dependencies:
     acorn "^4.0.4"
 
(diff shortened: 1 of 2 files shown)


dhei:
@guperrot yarn.lock should be checked into source control according to [Yarn docs](https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control):
> All yarn.lock files should be checked into source control (e.g. git or mercurial). This allows Yarn to install the same exact dependency tree across all machines, whether it be your coworker’s laptop or a CI server.

PR #136, opened 2017-10-11
Implement Android bridge for RN crash changes

guperrot:
Plus minor fixes.


mobile-center-crashes/Crashes.js
```diff
@@ -1,21 +1,23 @@
...
 
 // This is set later if and when the user provides a value for the getErrorAttachments callback
-let getErrorAttachmentsMethod = () => {};
+let getErrorAttachmentsMethod = () => { };
 const filteredReports = [];
 
 const UserConfirmation = {
     Send: 1,
     DontSend: 2,
-    AlwaySend: 3
```
achocron: lol

mobile-center-crashes/android/src/main/java/com/microsoft/azure/mobile/react/crashes/RNCrashesModule.java
```diff
@@ -114,12 +111,58 @@ public void run() {
     }
 
     @ReactMethod
-    public void crashUserResponse(boolean send, ReadableMap attachments, Promise promise) {
-        int response = send ? Crashes.SEND : Crashes.DONT_SEND;
-        if (mCrashListener != null) {
-            mCrashListener.setAttachments(attachments);
+    public void notifyWithUserConfirmation(int userConfirmation) {
+        Crashes.notifyUserConfirmation(userConfirmation - 1);
```
achocron: Wouldn't it be safer to use a switch statement? In the future, if we change the numbers for whatever reason, it would be such a pain to find every time we use them in wrapper sdks and update them

dhei: This `userConfirmation - 1` is really confusing without understanding Android SDK implementation. In Android SDK, we use 0 for `SEND`, 1 for `DONT_SEND` and 2 for `ALWAYS_SEND`, but in RN SDK we use 1 for `Send`, 2 for `DontSend` and 3 for `AlwaysSend`.

What's the reason we can't use `{Send: 0, DontSend: 1, AlwaysSend: 2}` for RN SDK? Can we keep them consistent?

mobile-center-crashes/Crashes.js
```diff
@@ -76,29 +81,31 @@ let Crashes = {
...
+                        if (!listenerMap.shouldProcess ||
+                            listenerMap.shouldProcess(report)) {
+                            filteredReports.push(report);
+                            filteredReportIds.push(report.id);
+                        }
+                    });
+                    RNCrashes.sendCrashReportsOrAwaitUserConfirmationForFilteredIds(filteredReportIds).then((alwaysSend) => {
+                        if (alwaysSend) {
+                            Helper.sendErrorAttachments(filteredReports);
+                        } else if (!listenerMap.shouldAwaitUserConfirmation || !listenerMap.shouldAwaitUserConfirmation()) {
```
dhei: not sure why you need `!listenerMap.shouldAwaitUserConfirmation` here?

guperrot: Because the callback can be undefined since it's optional to define it.

mobile-center-crashes/android/src/main/java/com/microsoft/azure/mobile/react/crashes/RNCrashesModule.java
guperrot: iOS is inconsitent with Android already.

mobile-center-crashes/Crashes.js
```diff
@@ -108,14 +115,9 @@ let Helper = {
     }
 };
 
-// Android does not have "isDebuggerAttached" method
-if (Crashes && RNCrashes && RNCrashes.isDebuggerAttached) {
```
dhei: Can you explain why removing this code?

guperrot: It's not an API we expose or document. It's in portable but works only for RN, and is not even consistent with iOS.

guperrot: (The goal was to align APIs)

mobile-center-crashes/Crashes.js
dhei: Can you document this subtlety? It's not very obvious.

mobile-center-crashes/android/src/main/java/com/microsoft/azure/mobile/react/crashes/RNCrashesModule.java
```diff
@@ -114,12 +111,58 @@ public void run() {
     }
 
     @ReactMethod
-    public void crashUserResponse(boolean send, ReadableMap attachments, Promise promise) {
-        int response = send ? Crashes.SEND : Crashes.DONT_SEND;
-        if (mCrashListener != null) {
-            mCrashListener.setAttachments(attachments);
+    public void notifyWithUserConfirmation(int userConfirmation) {
```
dhei: Can we rename this method to `notifyUserConfirmation` to keep it consistent with RN Java implementation? We try to keep JS and Java API consistent in RN SDK in most places.

guperrot: This calls an iOS method so we can't really mess with iOS conventions here I guess.

mobile-center-crashes/Crashes.js
```diff
@@ -53,9 +56,11 @@ let Crashes = {
         return RNCrashes.setEnabled(shouldEnable);
     },
 
-    notifyWithUserConfirmation(userConfirmation) {
+    notifyUserConfirmation(userConfirmation) {
         RNCrashes.notifyWithUserConfirmation(userConfirmation);
-        Helper.sendErrorAttachments(getErrorAttachmentsMethod, filteredReports);
+        if (userConfirmation !== UserConfirmation.DontSend) {
```
dhei: Because `userConfirmation` is external user input and it can be any value or `undefined`, can we validate this value just to be safe?

guperrot: @dhei please create stories then as native SDKs don't check and they are the ones they should check.

mobile-center-crashes/android/src/main/java/com/microsoft/azure/mobile/react/crashes/RNCrashesModule.java
dhei: ok, make sense 👍

mobile-center-crashes/Crashes.js
guperrot: We do that a lot for every callback, it's very common for JS callbacks.

PR #171, opened 2017-11-21
Fix/remove yarn lock demo app and release script improvement

dhei:
Delete yarn.lock in #170. App Center Build will use `yarn install` when `yarn.lock` file is present in the repo. Yarn doesn't work well with private VSTS npm registry. So this fix will let App Center Build fallback to `npm install` for now.

PS: I also parameterize the script for packaging RN iOS framework.


AppCenterReactNativeShared/zip-framework.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 cd Products
-zipfilename=AppCenter-SDK-ReactNative-iOS-Pod-1.x.x.zip
+wrapperSdkStringn=$(grep s.version AppCenterReactNativeShared.podspec)
```
guperrot: typo?

dhei: good catch, thanks @guperrot

PR #177, opened 2017-12-01
Support linking brownfield apps

### LINKED ISSUE #94: Support use without `react-native link`; add `.podspec` for libraries
I have a so-called "brownfield" app; a native iOS Swift app with integrated React Native views.

To do this I had to add all React libraries as CocoaPods dependencies. But now when I add `mobile-center` as npm dependency and run `react-native link`, it's added correctly, but the project doesn't build because the linked project can't find the React library files. 

Up until now for all other libraries installed through npm, there was always a `podspec` file for the React Library too. So I just referred to the local podspec file in the node_modules directory, and didn't have to use `react-native link`.

So my question is; can there be podspecs added to the different libraries? Or would you accept a PR that does so?

dhei:
Starting from [react-native@0.50.0](https://github.com/facebook/react-native/releases/tag/v0.50.0), `react-native link` supports installing from `Podfile` for brownfield react native applications.

> This change will allow to link projects that contains its own .podspec file to CocoaPods-based projects. In case link detect Podfile in iOS directory, it will look for related .podspec file in linked project directory, and add it to Podfile. If Podfile and .podspec files are not present, it will fall back to previous implementation.

Previously it's difficult to install appcenter dependencies from Podfile via `pod install`. Even with `pod install` installed appcenter dependencies, users still need to figure out how to run rnpm postinstall script to configure appcenter-analytics and appcenter-crashes module correctly. With `react-native link` supports `Podfile` out-of-box, it's much easier to use our sdk in brownfield applications.

This should fix #94.

### DIFF
--- a/TestApp/ios/BrownfieldPodfileExample
+++ b/TestApp/ios/BrownfieldPodfileExample
@@ -1,6 +1,7 @@
 # Example Podfile for brownfield react native application
 # For more information about using `react-native link` install pods for brownfield application, 
 # please refer to https://github.com/facebook/react-native/commit/74146cb3158273921b768c97f229695e680ea82d
+# Note that this scenario is only supported in react-native v0.50.0 or later (https://github.com/facebook/react-native/releases/tag/v0.50.0)
 
 target 'TestApp' do
   # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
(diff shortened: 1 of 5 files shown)


dhei:
@guperrot A sample brownfield app is added in the most recent commit. 

There's no change in linker script because all the magic is done by react-native@0.50.0 itself. As long as brownfield iOS app add appcenter reference to `Podfile`, `react-native link` should take care of the rest.

PR #181, opened 2017-12-06
Process crash reports only if Crashes is enabled

bmourat:


appcenter-crashes/Crashes.js
```diff
@@ -93,26 +93,31 @@ const Crashes = {
...
-                    AppCenterReactNativeCrashes.sendCrashReportsOrAwaitUserConfirmationForFilteredIds(filteredReportIds).then((alwaysSend) => {
-                        if (alwaysSend) {
-                            Helper.sendErrorAttachments();
-                        } else if (!listenerMap.shouldAwaitUserConfirmation || !listenerMap.shouldAwaitUserConfirmation()) {
-                            Crashes.notifyUserConfirmation(UserConfirmation.SEND);
-                        }
-                    });
-                }
-            });
+        Crashes.isEnabled()
```
dhei: current implementation use nested Promises which is a well-known anti-pattern in JS. The problem with nested Promises is mainly with uncaught error handling and rejected Promises. 

For instance, the following code would solve the nested Promises probme:
```
        Crashes.isEnabled()
        .then((isEnabled) => {
            if (isEnabled) {
                return AppCenterReactNativeCrashes.getUnprocessedCrashReports();
            }
            throw new Error('Crashes service is not enabled.');
        }).then((reports) => {
            if (reports.length > 0) {
                const filteredReportIds = [];
                reports.forEach((report) => {
                    if (!listenerMap.shouldProcess ||
                        listenerMap.shouldProcess(report)) {
                        filteredReports.push(report);
                        filteredReportIds.push(report.id);
                    }
                });
                AppCenterReactNativeCrashes.sendCrashReportsOrAwaitUserConfirmationForFilteredIds(filteredReportIds).then((alwaysSend) => {
                    if (alwaysSend) {
                        Helper.sendErrorAttachments();
                    } else if (!listenerMap.shouldAwaitUserConfirmation || !listenerMap.shouldAwaitUserConfirmation()) {
                        Crashes.notifyUserConfirmation(UserConfirmation.SEND);
                    }
                });
            }
        }).catch(error => console.log(error));
```

PR #193, opened 2017-12-13
Do not show error in test app when binary attachment is not present

bmourat:


TestApp/AttachmentsProvider.js
```diff
@@ -31,7 +31,7 @@ export default class AttachmentsProvider {
     try {
       contents = await RNFS.readFile(path, DEFAULT_ENCODING);
     } catch (error) {
-      console.error('Error while reading binary attachment file');
+      console.log('Error while reading binary attachment file');
```
dhei: console.log(`Error while reading binary attachment file, error: ${error}`);

TestApp/AttachmentsProvider.js
```diff
@@ -31,7 +31,7 @@ export default class AttachmentsProvider {
     try {
       contents = await RNFS.readFile(path, DEFAULT_ENCODING);
     } catch (error) {
-      console.error('Error while reading binary attachment file');
+      console.log(`Error while reading binary attachment file, error: ${error}`);
```
dhei: Can you make the same change for DemoApp/AttachmentsProvider.js ?

Issue #201, opened 2017-12-19
Analytics.trackEvent doesn't allow message with no props


raymondchooi:
### Expected behaviour

`Analytics.trackEvent("Log in successful.");` should run without errors

### Actual behaviour

Error: 
`undefined is not an object (evaluating 'Object.keys(props)') `
in Analytics.js:26:16

### Steps to reproduce

- Add analytics following https://docs.microsoft.com/en-us/appcenter/sdk/analytics/react-native
- Track an event with just a message, e.g.
`Analytics.trackEvent("Log in successful.");`

### Possible solution

Add default props to equal null and add early break in Analytics.js:20:
```
function sanitizeProperties(props = null) {
    if(props === null) {
        return {};
    }
....
```

dhei:
@raymondchooi thanks for reporting.👍  
We will fix this in our next release.

PR #202, opened 2017-12-19
Allow omitting properties in trackEvent call

bmourat:
This addresses #201


dhei:
@bmourat This change looks good. Can you add a UI test to `AnalyticsScreen.js` of TestApp & TestApp34 like this:

```javascript
<TouchableOpacity onPress={() => Analytics.trackEvent('Event without properties')}>
  <Text style={SharedStyles.button}>
    Track Event without properties
  </Text>
</TouchableOpacity>
```

PR #217, opened 2018-01-14
Let the script only check if the pod binary is available

francescopersico:
From my experience there is no need to check the platform. If the `pod` binary is available the dependencies will be sucessfully added and installed. I am on a Linux machine, my project uses rbenv and cocoapods, and all is working when i try to link the libraries.

### DIFF
--- a/appcenter-link-scripts/src/ios/index.js
+++ b/appcenter-link-scripts/src/ios/index.js
@@ -73,9 +73,6 @@ module.exports = {
     },
 
     addPodDeps(pods) {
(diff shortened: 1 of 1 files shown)


guperrot:
Hi and thanks for your contribution.
My guess is that this check was to make the error meaningful mostly on Windows but I guess the error check that follows right is good enough. @dhei I let you merge if you are ok.

dhei:
@francescopersico Thanks for your PR. 

This is an interesting scenario. The platform check is for leaving meaningful error message when running `react-native link` on Windows.

PR #243, opened 2018-02-16
Update demo app

achocron:


DemoApp/ios/DemoApp/AppDelegate.m
```diff
@@ -25,7 +25,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
   [AppCenterReactNativePush register];  // Initialize AppCenter push
 
-  [AppCenterReactNativeCrashes registerWithAutomaticProcessing];  // Initialize AppCenter crashes
+  [AppCenterReactNativeCrashes register];  // Initialize AppCenter crashes
```
guperrot: Is register manual processing? Really those ios methods are poorly named I am always confused xD

achocron: I think so? this wasn't a manual change

guperrot: Did you test attachments? and user confirmation?

achocron: I was able to send attachments

achocron: and i got the user confirmation pop up

guperrot: then @TroubleMakerBen did not test that last time as automatic process was enabled before, no way it could have worked

Issue #172, opened 2017-11-23
Add typescript declaration files


KSanthanam:
The Appcenter modules for react-native are missing the declaration (d.ts) files. Can you please add them.

lucamorelli:
seems I have same problem with the cordova plugin

guperrot:
Hi the cordova plugin is managed in another repository: https://github.com/Microsoft/AppCenter-SDK-Cordova, please create cordova issues there.

guperrot:
Typescript files are available on RN SDK, we forgot to close this one.

PR #247, opened 2018-02-23
Add typescript declaration files for all public APIs

janpieterz:
Let me know if this is appreciated, I can add types for the other packages too.


guperrot:
(Subset of #172)

appcenter-analytics/Analytics.d.ts
```diff
@@ -0,0 +1,6 @@
+export function trackEvent(
+  eventName: string,
+  properties?: { [name: string]: string | number | boolean }
```
guperrot: We support only string property values

appcenter-analytics/Analytics.d.ts
```diff
@@ -0,0 +1,6 @@
...
+): Promise<string | null>;
+export function isEnabled(): Promise<boolean | string>;
```
guperrot: The promise is boolean only, not string

appcenter-analytics/Analytics.d.ts
```diff
@@ -0,0 +1,6 @@
...
+export function setEnabled(enabled: boolean): Promise<string | null>;
```
guperrot: The promise does not return a string.

elamalani:
@janpieterz Thanks for submitting a PR in our SDK and we really appreciate your contribution. Our team reviewed your PR and it looks like you are adding typescript definition only in Analytics module. For this to be consistent across, all the modules - appcenter, appcenter-push, appcenter-crashes should have the definitions as well.

guperrot:
We would also need to make a Typescript version of TestApp before merging to validate the files.

appcenter-analytics/Analytics.d.ts
janpieterz: I thought so but then looked at the `sanitizeProperties` function and saw it parse strings, numbers and booleans too I think?

appcenter-analytics/Analytics.d.ts
janpieterz: I read https://docs.microsoft.com/en-us/appcenter/sdk/analytics/react-native#check-if-app-center-analytics-is-enabled as that it'd return a string (error) if it fails. Does it throw an error if it fails?

appcenter-analytics/Analytics.d.ts
janpieterz: Would you prefer setting it to an empty return then `: Promise` or `: Promise<null>`? Returning null seems like an odd thing to specify.

appcenter-analytics/Analytics.d.ts
guperrot: This API cannot fail, its asynchronous but just a getter for a boolean value.

appcenter-analytics/Analytics.d.ts
guperrot: I am not familiar with Promise typescript, but the type is "void", if Promise without type works, then yes we should use it.

appcenter-analytics/Analytics.d.ts
guperrot: We have custom properties for push which are typed and not using a dictionary anyway to avoid errors.

Custom properties and event properties are 2 different features. Event properties are just string values.

MatkovIvan: @guperrot This confusion caused by [this conversion to string](https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/appcenter-analytics/Analytics.js#L31-L35). And if it already here I guess it's OK.

guperrot: The only reason we have this is because object was used to map strings because javascript is not strongly typed. But in all typed SDKs we have it's really string values that we are allowing and sending a number would be converted to a string anyway. Exposing more types will be misleading.

janpieterz: 👍 Should be fixed. This one does return a `string | null`? There seems a lot more opportunity for errors here, but the docs mention same as with `isEnabled` (`This function returns a promise, in which the success case returns an empty string, and the error case returns the error.`)

appcenter-analytics/Analytics.d.ts
janpieterz: 👍 fixed
Might be worth adjusting the docs for this too, made a pull request there (https://github.com/MicrosoftDocs/appcenter-docs/pull/37).

appcenter-analytics/Analytics.d.ts
janpieterz: `Promise<void>` will do the trick 😄

janpieterz:
I'll tackle the `appcenter`, `appcenter-push`, `appcenter-crashes` as well, and create a typescript version of the `TestApp`. Might take a bit.

Would you prefer me tackling those all at once or pushing the definitions out per package so we keep the stuff to review small?

guperrot:
If we merge 1 by 1 then the target branch cannot be develop and we can create another hosting branch. Github offers incremental reviews since last time so it's not a problem for me as long as commits for TestApp are separated from the SDK changes it should be ok.

guperrot:
Changes on analytics are ok so far.

janpieterz:
Should have all typings now, let me know what's wrong and I'll fix it, then move on to the TestApp.

appcenter-crashes/Crashes.d.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export interface ListenerMap {
+  onBeforeSending?: () => void;
+  onSendingSucceeded?: () => void;
+  onSendingFailed?: () => void;
+  getErrorAttachments?: (report) => Promise<[TextAttachment, FileAttachment]>;
+}
+
+interface TextAttachment {
```
guperrot: Text and File attachment are the same ErrorAttachmentLog class on android and .net SDKs, please stick to other SDK conventions for now. The methods are actually statics that return the same type.

guperrot: @guperrot
guperrot just now  Member
Text and File attachment are the same ErrorAttachmentLog class on android and .net SDKs, please stick to other SDK conventions for now. The methods are actually statics that return the same type.

appcenter-crashes/Crashes.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    data: object,
+    fileName: string,
+    contentType: string
+  ): FileAttachment;
+}
+
+export default interface Crashes {
+  generateTestCrash(): Promise<void>;
+  hasCrashedInLastSession(): Promise<boolean>;
+  lastSessionCrashReport(): Promise<X>;
```
guperrot: what is X?

appcenter-push/Push.d.ts
```diff
@@ -0,0 +1,13 @@
+export interface PushNotification {
+    message?: string | null;
+    title?: string | null;
+    customProperties?: { [name: string]: stirng };
```
guperrot: Typo

appcenter-push/Push.d.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export interface ListenerMap {
```
guperrot: Please use our Android SDK name convention.

appcenter/AppCenter.d.ts
```diff
@@ -0,0 +1,21 @@
...
+    LogLevelDebug = 3,
+    LogLevelInfo = 4,
+    LogLevelWarning = 5,
+    LogLevelError = 6,
+    LogLevelAssert = 7,
+    LogLevelNone = 99
+}
+
+export interface CustomProperties {
+    set(key: string, value: string);
```
guperrot: Please look at our docs and also Android SDK, we have multiple types for this one.

appcenter-push/Push.d.ts
janpieterz: Just done the typescript definitions, not the backing js (`setListener(listenerMap)`)

appcenter-crashes/Crashes.d.ts
janpieterz: It's an ErrorReport. Not too familiar with how this works behind the scenes. Looks almost like it's a dictionary/map behind the scenes in both android (https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/appcenter-crashes/android/src/main/java/com/microsoft/appcenter/crashes/AppCenterReactNativeCrashesUtils.java) and ios (https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/appcenter-crashes/ios/AppCenterReactNativeCrashes/AppCenterReactNativeCrashesUtils.m).

Let me know if attempt (coming up soon) is correct

appcenter/AppCenter.d.ts
janpieterz: Looks like a simple map on both, let me know if this works.

appcenter-crashes/Crashes.d.ts
```diff
@@ -0,0 +1,60 @@
+export const enum UserConfirmation {
+  DONT_SEND = 0,
+  SEND = 1,
+  ALWAYS_SEND = 2
+}
+
+export interface ListenerMap {
```
guperrot: CrashesListener is used in Android SDK.

appcenter-crashes/Crashes.d.ts
```diff
@@ -0,0 +1,60 @@
...
+  appBuild: string;
+  appNamespace: string;
+}
+
+export interface ErrorReport {
+  id: string;
+  threadName: string;
+  appErrorTime: string;
+  appStartTime: string;
+  exception: string;
```
guperrot: can you double check type here?

janpieterz: `errorReportMap.putString("exception", Log.getStackTraceString(errorReport.getThrowable()));`
`errorReportMap.putString("exceptionReason", errorReport.getThrowable().getMessage());`

Both in https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/appcenter-crashes/android/src/main/java/com/microsoft/appcenter/crashes/AppCenterReactNativeCrashesUtils.java#L32 which is called from https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/appcenter-crashes/android/src/main/java/com/microsoft/appcenter/crashes/AppCenterReactNativeCrashesModule.java#L77

iOS same story: https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/appcenter-crashes/ios/AppCenterReactNativeCrashes/AppCenterReactNativeCrashesUtils.m#L110 called from https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/appcenter-crashes/ios/AppCenterReactNativeCrashes/AppCenterReactNativeCrashes.m#L111

guperrot:
Looks good, now we would need a clone of the testapp we have in the repo but in typescript version to validate all the changes.

janpieterz:
Great stuff, will start tackling that.

janpieterz:
@guperrot Made some progress with the test app.

Couple of questions:

1. PushNotification title can currently be a string, null or undefined. It looks like the test app doesn't account for it possibly being undefined. Is this something to change in the Typescript definition? Is it only null/undefined if message is null/undefined?
1. ErrorAttachmentLog.attachmentWithBinary seems to be called with a string from the test apps, is this always the case?
1. Did I miss `shouldProcess` and `shouldAwaitUserConfirmation` in the CrashesListener interface before? I've added them now, just making sure it's not a private API and it's supposed to be public.
1. On analyticsScreen a variable data is used, undefined (I think). It's for `Track Event badly (Do not do this, only strings are supported)`. What is supposed to happen here? It currently doesn't do anything.

TestAppTypescript/android/app/src/main/java/com/testapptypescript/MainApplication.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Override
+  public ReactNativeHost getReactNativeHost() {
+    return mReactNativeHost;
+  }
+
+  @Override
+  public void onCreate() {
+    super.onCreate();
+    SoLoader.init(this, /* native exopackage */ false);
```
guperrot: Please copy the code from original test app here, missing the sender id and verbose logs.

TestAppTypescript/android/app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,5 @@
+<resources>
+    <string moduleConfig="true" name="appcenterCrashes_whenToSendCrashes">ALWAYS_SEND</string>
```
guperrot: Use same settings as original test app otherwise crash listener won't work.

TestAppTypescript/ios/TestAppTypescript/AppDelegate.m
```diff
@@ -0,0 +1,49 @@
...
+#import <React/RCTBundleURLProvider.h>
+#import <React/RCTRootView.h>
+
+@implementation AppDelegate
+
+- (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
+{
+  NSURL *jsCodeLocation;
+
+  [AppCenterReactNativePush register];  // Initialize AppCenter push
```
guperrot: Same comment as Android to report manual changes we did in TestApp.

TestAppTypescript/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "start": "node node_modules/react-native/local-cli/cli.js start",
+    "tsc": "tsc",
+    "watch": "tsc -w",
+    "ios": "npm run build && concurrently -r \"npm run watch\" \"react-native run-ios\"",
+    "android": "npm run build && concurrently -r \"start npm run watch\" \"start react-native run-android\"",
+    "build": "npm run clean && npm run tsc",
+    "clean": "rimraf artifacts"
+  },
+  "dependencies": {
+    "appcenter": "file:appcenter-1.3.0.tgz",
```
guperrot: We are using 1.4.0, can you catch up with develop?

guperrot:
https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/247#issuecomment-378589419

1) Sounds like a bug, it should be string or null but not undefined. We will have to investigate.
2) From Javascript we pass binary as string with base64 encoding, only native SDK APIs have support for byte arrays.
3) Those APIs are not new and are [publicly documented](https://docs.microsoft.com/en-us/appcenter/sdk/crashes/react-native#customize-your-usage-of-app-center-crashes).
4) I have to check but AppCenter is supposed to log an error in the console and not send the event. In TypeScript we have types so you don't need that button I guess.

I made a first quick review for changes on TestApp and need some fixes before I can deeply review and test.

guperrot:
For Android its either string or null for push notification title and message, never undefined.
I have to test iOS.

guperrot:
Undefined comes from iOS implementation, something that we should fix.

guperrot:
With #277 it will be null or string but never undefined, you can reflect that on the types ;)

janpieterz:
Will have to loop back here another time for getting 1.4.0 in it, the image picking stuff, confirming ios/android builds.

appcenter-push/Push.d.ts
```diff
@@ -1,6 +1,6 @@
 export interface PushNotification {
-    message?: string | null;
-    title?: string | null;
+    message: string | null;
+    title: string | null;
     customProperties?: { [name: string]: string };
```
guperrot: is the question mark meaning undefined? If yes then customProperties is actually always defined and empty map if no custom property.

janpieterz: Question mark means optional, in effect it is undefined when not set. 
Technically the better thing here would've been: `string | null | undefined` since it's never created by a library consumer. 

Anyway, I'll remove the question from customProperties, cheers!

janpieterz:
@guperrot suggestion for how to best merge latest `develop` branch? Can I just add this as a remote and pull it in? Been trying some stuff but that didn't work great.

guperrot:
```
git remote add upstream https://github.com/Microsoft/AppCenter-SDK-React-Native
git fetch upstream # every time
git merge upstream/develop
```

guperrot:
@janpieterz The following should work:

```
git remote add upstream https://github.com/Microsoft/AppCenter-SDK-React-Native.git

git fetch upstream # every time
git merge upstream/develop
```

TestAppTypescript/update-link-to-appcenter.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
guperrot: Make both scripts executable

TestAppTypescript/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "babel-jest": "22.4.3",
+    "babel-preset-react-native": "4.0.0",
+    "concurrently": "3.5.1",
+    "prettier": "1.10.2",
+    "rimraf": "2.6.2"
+  },
+  "jest": {
+    "preset": "react-native"
+  }
+}
```
guperrot: add end of line, my vs code is trying to add those and thus generate a git diff
same things for the other new package.json files.

TestAppTypescript/update-link-to-appcenter.sh
guperrot: gradlew too

TestAppTypescript/artifacts/AppCenterScreen.js
```diff
@@ -0,0 +1,126 @@
...
+                .set('now', new Date());
+            yield AppCenter.setCustomProperties(properties);
+        });
+    }
+    render() {
+        return (<View style={SharedStyles.container}>
+                <ScrollView>
+                    <Text style={SharedStyles.heading}>
+                        Test AppCenter
+          </Text>
```
guperrot: Fix formatting in all files.

TestAppTypescript/src/AnalyticsScreen.tsx
```diff
@@ -0,0 +1,89 @@
...
+                    </TouchableOpacity>
+
+                    <TouchableOpacity onPress={() => Analytics.trackEvent('Button press', { propertyValueTooLong: '12345678901234567890123456789012345678901234567890123456789012345' })}>
+                        <Text style={SharedStyles.button}>
+                            Track Event - event property value truncated after 64 characters
+            </Text>
+                    </TouchableOpacity>
+                    <TouchableOpacity onPress={() => Analytics.trackEvent('Button press', undefined)}>
+                        <Text style={SharedStyles.button}>
+                            Track Event badly (Do not do this, only strings are supported)
```
guperrot: Remove that specific button for the typescript app as in typescript it's not possible to the same as original TestApp, as discussed previously.

TestAppTypescript/src/MainScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+      </View>
+    );
+  }
+}
+
+Push.setListener({
+  onPushNotificationReceived(pushNotification) {
+    let message = pushNotification.message;
+    let title = pushNotification.title;
+    //TODO verify title undefined/null
```
guperrot: Implement the todo since merged with develop that has the fix now.
Please look at TestApp updated code actually as we don't mention android anymore in the message.

dhei:
Hey @janpieterz, it has been a while. Just checking with you, do you have a chance to address @guperrot's comments?

dhei:
@janpieterz, Congratulations! ❤️ 
I am looking forward to merging this PR.

TestAppTypescript/artifacts/AppCenterScreen.js
janpieterz: Artifacts shouldn't have been committed. Will remove.

janpieterz: Have formatted other files with Prettier.

TestAppTypescript/src/AnalyticsScreen.tsx
janpieterz: Done

TestAppTypescript/src/MainScreen.tsx
janpieterz: Not 100% sure on this, can you check upcoming commit?

TestAppTypescript/package.json
janpieterz: Done. Might need a committed .vscode to the projects to ensure no-one bumps into this?

janpieterz:
I'll loop back this afternoon (hopefully), as I don't have a MacOS machine close (nor often work on it).

I'll also pull in changes of 1.6.0. Just checked all changelogs, and there doesn't seem to be any API changes, right?

dhei:
Hey @janpieterz,

Thanks for the update.

Looks like this PR is created when this sdk is at v1.4.0. There has been no API change since v1.4.0, as you can see the release notes of [v1.5.0](https://github.com/Microsoft/AppCenter-SDK-React-Native/releases/tag/1.5.0), [v1.5.1](https://github.com/Microsoft/AppCenter-SDK-React-Native/releases/tag/1.5.1), [v1.6.0](https://github.com/Microsoft/AppCenter-SDK-React-Native/releases/tag/1.6.0).

janpieterz:
Having some issues finding time with a Mac machine.

andreialecu:
I find it interesting that a Microsoft project does not have TypeScript definitions. Shouldn't this be a company policy? :)

janpieterz:
@guperrot / @dhei updated to work with iOS and latest (1.8.1) version. Image picker fixed as well.

Still needs testing with an appcenter account of course.

.gitignore
```diff
@@ -45,6 +46,10 @@ android/keystores/debug.keystore
 
 # credentials
 .npmrc
+google-services.json
```
guperrot: We found a way to commit these files without secret so you can remove this line.
Please refer to https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/TestApp/android/app/google-services.json if you use the same package name in the ts test app then you can just use the same file.

TestAppTypescript/android/app/src/main/java/com/testapptypescript/MainApplication.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public boolean getUseDeveloperSupport() {
+      return BuildConfig.DEBUG;
+    }
+
+    @Override
+    protected List<ReactPackage> getPackages() {
+      return Arrays.<ReactPackage>asList(
+          new MainReactPackage(),
+            new AppCenterReactNativePushPackage(MainApplication.this),
```
guperrot: Duplicated linked modules here

TestAppTypescript/android/app/src/main/res/values/strings.xml
```diff
@@ -0,0 +1,7 @@
+<resources>
+    <string moduleConfig="true" name="appcenterCrashes_whenToSendCrashes">ASK_JAVASCRIPT</string>
```
guperrot: duplicates should be removed here as well

package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
```
guperrot: We have this file in git ignore even though general practice checks this in. We are managing packages dynamically in our apps. Please delete.

.gitignore
janpieterz: Weird that it didn't get merged. What is the google-services used for, and how do I make sure that works besides just adding it as a build dependency?

TestAppTypescript/android/app/src/main/java/com/testapptypescript/MainApplication.java
janpieterz: Very odd, didn't touch this manually, maybe an error in the linking?

.gitignore
guperrot: Like in TestApp, make sure firebase is integrated: https://docs.microsoft.com/en-us/appcenter/sdk/push/migration/react-native-android. 

When the procedure asks to download google-services.json, pick https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/TestApp/android/app/google-services.json instead and make sure your app has the same package name as TestApp.

TestAppTypescript/android/app/src/main/java/com/testapptypescript/MainApplication.java
guperrot: We fixed that recently but maybe it was run before.

janpieterz:
@guperrot Google Services added, not sure what else there is atm.

FutureGUIs:
I'm not sure it's truly the fix for that latest 'bundling' error, but I used to get it when just running the normal react-native android or ios on RN 0.54 or 0.55.  But then I would run a separate `npm start`  or the RN `\cli.js start` and I'd be able to run my app.   And after upgrading to the new RN 0.57 release, that issue has gone away.

guperrot:
I already tried npm run start as separate but it didn't change anything, same error.

janpieterz:
@guperrot This is because you're not running the Typescript compiler first. You're welcome to change the commands to something more to your liking, running `npm run watch` is needed.

There should be commands for `npm run ios` and `npm run android` that should automatically start the correct watcher and react-native run-X.

guperrot:
```
14:19 $ npm run android
npm ERR! missing script: android

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/guperrot/.npm/_logs/2018-09-13T21_19_57_260Z-debug.log
✘-1 ~/git/appcenter-sdk-react-native/TestApp [:bcc2b6b|⚑ 2] 
14:19 $ npm run watch
npm ERR! missing script: watch

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/guperrot/.npm/_logs/2018-09-13T21_20_04_119Z-debug.log

**Comment by guperrot** [comment]
`npm run watch` did the trick, I could run test app successfully on both platforms.
However I am investigating on typing inconsistencies in crash report, will elaborate soon but we might have to update either SDK or typings.

File: appcenter/AppCenter.d.ts [hunk 25]
```diff
@@ -0,0 +1,30 @@
+export const enum AppCenterLogLevel {
+    LogLevelVerbose = 2,
```
guperrot: Using pascal case for enums in our internal code was an inconsistency with Crashes.

In the typescript version since we are defining new const we should take the opportunity to align.

In other APIs we are adding we are planning to use THIS_CONVENTION.

Let's change this enum to using the following values without the LogLevel prefix as we are already in the enum:

```
   VERBOSE = 2,
   DEBUG = 3,
   // etc...
```

appcenter/AppCenter.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    LogLevelAssert = 7,
+    LogLevelNone = 99
+}
+
+export class CustomProperties {
+    set(key: string, value: string | number | boolean | Date);
+    clear(key: string);
+}
+
+export const LogLevelVerbose: AppCenterLogLevel.LogLevelVerbose;
```
guperrot: We have the enum, remove the top level constants.

appcenter/AppCenter.d.ts
guperrot: We should also reference `AppCenterLogLevel.LogLevelDebug` instead of using the integer value directly here, like you did for constants at top level.

TestAppTypescript/ios/TestAppTypescript/Info.plist
```diff
@@ -0,0 +1,64 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
guperrot: Please git apply this patch so that we can test with push, had to this locally:

```
diff --git a/TestAppTypescript/ios/TestAppTypescript/Info.plist b/TestAppTypescript/ios/TestAppTypescript/Info.plist
index d11cea8..fff4790 100644
--- a/TestAppTypescript/ios/TestAppTypescript/Info.plist
+++ b/TestAppTypescript/ios/TestAppTypescript/Info.plist
@@ -22,6 +22,19 @@
        <string>????</string>
        <key>CFBundleVersion</key>
        <string>1</string>
+       <key>LSRequiresIPhoneOS</key>
+       <true/>
+       <key>NSAppTransportSecurity</key>
+       <dict>
+               <key>NSExceptionDomains</key>
+               <dict>
+                       <key>localhost</key>
+                       <dict>
+                               <key>NSExceptionAllowsInsecureHTTPLoads</key>
+                               <true/>
+                       </dict>
+               </dict>
+       </dict>
        <key>NSCameraUsageDescription</key>
        <string>$(PRODUCT_NAME) would like to use your camera</string>
        <key>NSLocationWhenInUseUsageDescription</key>
@@ -30,8 +43,10 @@
        <string>$(PRODUCT_NAME) would like to save photos to your photo gallery</string>
        <key>NSPhotoLibraryUsageDescription</key>
        <string>$(PRODUCT_NAME) would like access to your photo gallery</string>
-       <key>LSRequiresIPhoneOS</key>
-       <true/>
+       <key>UIBackgroundModes</key>
+       <array>
+               <string>remote-notification</string>
+       </array>
        <key>UILaunchStoryboardName</key>
        <string>LaunchScreen</string>
        <key>UIRequiredDeviceCapabilities</key>
@@ -46,19 +61,5 @@
        </array>
        <key>UIViewControllerBasedStatusBarAppearance</key>
        <false/>
-       <key>NSLocationWhenInUseUsageDescription</key>
-       <string></string>
-       <key>NSAppTransportSecurity</key>
-       <!--See http://ste.vn/2015/06/10/configuring-app-transport-security-ios-9-osx-10-11/ -->
-       <dict>
-               <key>NSExceptionDomains</key>
-               <dict>
-                       <key>localhost</key>
-                       <dict>
-                               <key>NSExceptionAllowsInsecureHTTPLoads</key>
-                               <true/>
-                       </dict>
-               </dict>
-       </dict>
 </dict>
 </plist>
```

TestAppTypescript/ios/TestAppTypescript.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,1923 @@
+// !$*UTF8*$!
```
guperrot: Please apply this patch for push/codesign for our team: 

[project.pbxproj.txt](https://github.com/Microsoft/AppCenter-SDK-React-Native/files/2381312/project.pbxproj.txt)

appcenter-crashes/Crashes.d.ts
```diff
@@ -0,0 +1,61 @@
...
+  getErrorAttachments?: (report: ErrorReport) => Promise<ErrorAttachmentLog[]>;
+  shouldProcess?: (report: ErrorReport) => boolean;
+  shouldAwaitUserConfirmation?: (report: ErrorReport) => boolean;
+}
+
+export class ErrorAttachmentLog {
+  public static attachmentWithText(text: string, fileName: string): ErrorAttachmentLog;
+  public static attachmentWithBinary(
+    data: string,
+    fileName: string,
```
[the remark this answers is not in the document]
janpieterz: In Typescript you can't have an optional parameter (?) followed by a required one (contentType).
Is content type also optional? Is it either only data or data + filename + contentType?

appcenter/AppCenter.d.ts
janpieterz: Constants at top level referenced the Typescript implementation (integers). They had to be changed to new CAPS method. I'm not sure we can reference a JS enum in a definition file, tried some ways but didn't work as it has to be a constant expression, any suggestions?

TestAppTypescript/ios/TestAppTypescript/Info.plist
janpieterz: Corrupt at the end when applying here. (Corrupt patch at line 57).

appcenter-crashes/Crashes.d.ts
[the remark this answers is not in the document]
guperrot: only fileName is optional, in that case use `string | null` instead of question mark.

appcenter/AppCenter.d.ts
guperrot: It's fine then, we can keep integers for now.

appcenter-analytics/Analytics.d.ts
```diff
@@ -1,6 +1,6 @@
 export function trackEvent(
   eventName: string,
   properties?: { [name: string]: string }
-): Promise<string | null>;
+): Promise<null>;
```
guperrot: You used void instead of null in other places, since we are actually returning undefined, it should not be null but void.

appcenter-analytics/Analytics.d.ts
```diff
@@ -0,0 +1,6 @@
+export function trackEvent(
+  eventName: string,
+  properties?: { [name: string]: string }
+): Promise<null>;
```
guperrot: Promis<void>

janpieterz:
@guperrot Should all be fixed.

guperrot:
@janpieterz thanks for this huge contribution 👍

dhei:
Hey @janpieterz, 

I want to say a big thank you! Your contribution is highly appreciated. 👍 

Thanks,
Di

janpieterz:
Thanks for the help gents, happy we landed it :)

PR #252, opened 2018-02-28
Fix react-native 0.53 prompt glitches when linking

guperrot:


appcenter-crashes/scripts/postlink.js
```diff
@@ -1,22 +1,53 @@
...
+                {
+                    name: 'Automatically',
+                    value: 'ALWAYS_SEND'
+                },
+                {
+                    name: 'Processed in JavaScript by user',
+                    value: 'ASK_JAVASCRIPT'
+                }]
+        }]).catch((e) => {
+            console.log(`Could not determine when to enable AppCenter crashes. Error Reason - ${e.message}`);
```
dhei: The original log message `Could not determine when to send AppCenter crashes.` sounds better to me.

appcenter-crashes/scripts/postlink.js
```diff
@@ -1,22 +1,53 @@
...
```
dhei: same as above comment.

appcenter-crashes/scripts/postlink.js
guperrot: Sounds like we didn't have the message as analytics then, I copied from there.

PR #256, opened 2018-03-02
Fix error catch logic in postlink scripts

guperrot:


appcenter-push/scripts/postlink.js
```diff
@@ -8,7 +8,7 @@ if (rnpmlink.android.checkIfAndroidDirectoryExists()) {
         .then(() => {
             rnpmlink.android.removeAndroidDuplicateLinks();
         }).catch((e) => {
-            console.log(`Could not configure AppCenter for Android. Error Reason - ${e.message}`);
```
dhei: `Could not configure AppCenter Push for Android`

PR #271, opened 2018-03-13
More refined on-boarding messaging.

pniko:
Cleaning up the messaging in the SDK so we can update the Getting Started page in the portal.

### DIFF
--- a/appcenter-link-scripts/src/android/index.js
+++ b/appcenter-link-scripts/src/android/index.js
@@ -30,7 +30,7 @@ module.exports = {
         }
         return inquirer.prompt([{
             type: 'input',
-            message: 'What is the Android App Secret?',
+            message: 'What secret does your Android app use? [None]',
             name: 'app_secret',
         }]).then((answers) => {
             config.set('app_secret', answers.app_secret);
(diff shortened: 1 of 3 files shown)


guperrot:
CI error was:

```

File: appcenter-push/scripts/postlink.js [hunk 1]
```diff
@@ -1,5 +1,8 @@
 const rnpmlink = require('appcenter-link-scripts');
 
+console.log("Please enter your Android and iOS app secrets below.");
```
dhei: please use single quotes ;)

appcenter-link-scripts/src/android/index.js
```diff
@@ -30,7 +30,7 @@ module.exports = {
         }
         return inquirer.prompt([{
             type: 'input',
-            message: 'What is the Android App Secret?',
+            message: 'What secret does your Android app use? [None]',
```
dhei: @pniko Question, so user has to delete `[None]` and then enter the app secret in the prompt?

guperrot: He is just changing the message, there is still no default value.

pniko: It's just to signal you can hit return. We are updating the Getting Started pages to stop directing users to flip to another app for the other secret. Narrowing the Getting Started instructions to the platform they are staring at makes things a bit cleaner until we let users group apps together in the future.

PR #292, opened 2018-04-25
add IDEWorkspaceChecks.plist to DemoApp

dhei:


dhei:
Once #295 is merged, the Eslint error should go away.

Issue #298, opened 2018-04-27
Jest tests fail when importing appcenter or appcenter-push


antwan2093:
### **Description**

Compatibility with Jest

Cannot test a connected container that includes `import AppCenter from 'appcenter';`

Jest throws an error which prevents further test execution.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. import AppCenter from 'appcenter'; in a react-native container component.
2. Write a snapshot test to test that component
3. Execute said test

### **Details**

1. Which SDK version are you using?
    - 1.5.0
2. Which OS version did you experience the issue on?
    - N/A
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - N/A
4. What third party libraries are you using?
    - Jest
5. Run the following command and paste the output below: `react-native info`

Place output here

6. If you're developing for React Native iOS, run the following command and paste the output below: `pod --version`

Environment:
  OS: Windows 10
  Node: 8.10.0
  Yarn: Not Found
  npm: 5.6.0
  Watchman: Not Found
  Xcode: N/A
  Android Studio: Version  3.0.0.0 AI-171.4443003

Packages: (wanted => installed)
  react: ^16.3.2 => 16.3.2
  react-native: 0.49.3 => 0.49.3

7. Please enable verbose logging for your app using `[MSAppCenter setLogLevel: MSLogLevelVerbose];` before your call to `[AppCenterReactNative register];` (or any other SDK registration). For Android, use `AppCenter.setLogLevel(android.util.Log.VERBOSE);` before your `SoLoader.init` call. Include the logs here:

N/A

From AppCenter-Push 
```Invariant Violation: Native module cannot be null.

      at invariant (node_modules/react-native/node_modules/fbjs/lib/invariant.js:44:15)
      at new NativeEventEmitter (node_modules/react-native/Libraries/EventEmitter/NativeEventEmitter.js:31:1)
      at Object.<anonymous> (node_modules/appcenter-push/Push.js:5:26)
      at Object.<anonymous> (logic/actions/settingsActions.js:39:128)
```

jmedellinMS:
Hi @antwan2093 ,
Thank you for reporting this. This looks like a bug and our team has an item in backlog for tracking this. We will update the thread once the fix is released. Sorry for the inconvenience.

dhei:
This should be fixed in [v1.6.0](https://github.com/Microsoft/AppCenter-SDK-React-Native/releases/tag/1.6.0).

PR #303, opened 2018-05-10
Create empty mocks for jest

bmourat:
Fixes second reason for #298. First was fixed here #301.
We could add this logic to postlink script but I think it makes more sense doing it after installation since it is not part of the linking process.


appcenter-crashes/scripts/postinstall.js
```diff
@@ -0,0 +1,50 @@
+const fs = require('fs');
+const path = require('path');
+
+const projectDirectory = path.resolve(`${__dirname}${path.sep}..${path.sep}..${path.sep}..`);
+const testDirectory = path.join(projectDirectory, 'test');
```
ajwhite: This assume's the dependent's directory structure, and overall feel a little uneasy about "injecting" source files into a dependent.

What I do find interesting/appealing is the automatic updating of `packageJson.jest.setupFiles`. 

How about defining the mock file _in this project_ and link to it, rather than creating it in the user's directory?

Eg:
```js
projectJson.jest.setupFiles.push('node_modules/appcenter-crashes/jest/mocks.js')
```

appcenter-crashes/scripts/postinstall.js
```diff
@@ -0,0 +1,50 @@
...
+    fs.writeFileSync(packageJsonFile, JSON.stringify(projectJson));
+}
+
+// Create setup mock file for Jest
+if (!fs.existsSync(testDirectory)) {
+    fs.mkdirSync(testDirectory);
+}
+
+fs.writeFileSync(`${testDirectory}/${setupFileName}`, `
+jest.mock('NativeModules', () => ({
```
ajwhite: How does `jest.mock` handle the scenario where the user has already defined a mock for `NativeModules`? Does this override/conflict with that? Or simply merge them?

appcenter-crashes/scripts/postinstall.js
```diff
@@ -0,0 +1,50 @@
...
+
+// Update project.json
+let packageJsonContent;
+try {
+    packageJsonContent = fs.readFileSync(packageJsonFile, 'utf8');
+} catch (e) {
+    console.log('Could not read package.json file');
+    return;
+}
+const projectJson = JSON.parse(packageJsonContent);
```
dhei: nit - is there a reason to name this variable "projectJson" not "packageJson"?

bmourat: Just a typo, thanks!

bmourat:
@ajwhite @dhei 
I think I found more elegant and less breakable solution. I decided not to mock NativeModules since there could be conflicts if a user has already defined mock for this module. I'm just creating mock files for each module in __mocks__ directory (the directory where jest expects to find mocks). Thus, there is no need to modify package.json file and user always can tweak our mocks as he wishes.

appcenter-crashes/scripts/postinstall.js
```diff
@@ -0,0 +1,37 @@
+const fs = require('fs');
+const path = require('path');
+
+const projectDirectory = path.resolve(__dirname, '..', '..', '..');
+const mockFileContent = `
```
dhei: nit - If you move the backtick(<code>`</code>) to the beginning of line 6, then the generated **__mocks__/appcenter-crashes.js** file won't have a empty new line at the beginning of the file.

appcenter-crashes/scripts/postinstall.js
```diff
@@ -0,0 +1,37 @@
...
+}
+
+// Create mock file for Jest
+const mocksDirectory = `${projectDirectory}/__mocks__`;
+const mockFileName = 'appcenter-crashes.js';
+if (!fs.existsSync(`${mocksDirectory}/${mockFileName}`)) {
+    if (!fs.existsSync(mocksDirectory)) {
+        fs.mkdirSync(mocksDirectory);
+    }
+    fs.writeFileSync(`${mocksDirectory}/${mockFileName}`, mockFileContent);
```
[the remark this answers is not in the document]
guperrot: @bmourat let's adopt the solution described here, I would vote for the first approach:

> Automatically add the file path to packageJson.jest.setupFiles during postinstall.

appcenter-push/scripts/postinstall.js
```diff
@@ -0,0 +1,29 @@
+const fs = require('fs');
+const path = require('path');
+
+const projectDirectory = path.resolve(__dirname, '..', '..', '..');
+const setupFileName = `.${path.sep}node_modules${path.sep}appcenter-push${path.sep}test${path.sep}appCenterPushMock.js`;
+
+// Check if package.json has jest as dependency
+let packageJson = '';
+const packageJsonFile = path.join(`${projectDirectory}`, 'package.json');
```
guperrot: This is a lot of duplicated code, we can share code in the linker scripts package and use it here.

guperrot:
@bmourat Work well for npm, however it failed when I tested with yarn. The cause is https://github.com/yarnpkg/yarn/issues/853 so will need to document the manual steps for Yarn users...

PR #314, opened 2018-05-17
Fix jest in TestApp

MatkovIvan:


dhei:
@MatkovIvan, eslint is not happy with mocks in `__mocks__` thus failing Build, please fix it.

guperrot:
@MatkovIvan Let's wait #303 to do that and have test app point at mock files provided by SDK.

TestApp/__mocks__/appcenter-crashes.js
```diff
@@ -0,0 +1,9 @@
+const Crashes = jest.mock('appcenter-crashes');
```
guperrot: Please catch up with develop and remove those. Might need to revisit the PR entirely now that #303 is merged.

guperrot:
When I run that I have the following:

```

PR #320, opened 2018-05-28
Update build tools to latest version

### LINKED ISSUE #318: Upgrade build tools SDK to 25
### **Description**

We have a React Native project that is using AppCenter SDKs. The build is quite slow, and we would like to upgrade Gradle to 3.3 and the Android plugin to 2.3.0. However, this requires all the dependencies in our project to be on at least build tools 25.
(See https://developer.android.com/studio/releases/gradle-plugin)

Since React Native projects must build the AppCenter projects as part of their build process, the AppCenter tools need to be upgraded.

Alternatively, is it supported to download built artifacts via Maven, the way a non-React Native Android project would?

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Edit `/android/build.gradle` and set the gradle plugin to `com.android.tools.build:gradle:2.3.3`
2. Edit `gradle-wrapper.properties` and change the Gradle version to 3.3.
3. Rebuild the project.

Gradle reports the following error:
```
FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring project ':app'.
> Could not resolve all dependencies for configuration ':app:_debugApk'.
   > A problem occurred configuring project ':appcenter-crashes'.
      > The SDK Build Tools revision (23.0.1) is too low for project ':appcenter-crashes'. Minimum required is 25.0.0
```

### **Details**

1. Which SDK version are you using?
    - 1.1.0
2. Which OS version did you experience the issue on?
    - N/A
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - N/A
4. What third party libraries are you using?
    - react-native-code-push
5. Run the following command and paste the output below: `react-native info`

```
Environment:
  OS: macOS High Sierra 10.13.4
  Node: 8.11.1
  Yarn: 1.5.1
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.3.1 Build version 9E501
  Android Studio: 2.3 AI-162.4069837

Packages: (wanted => installed)
  react: 16.0.0 => 16.0.0
  react-native: 0.50.4 => 0.50.4
```

6. If you're developing for React Native iOS, run the following command and paste the output below: `pod --version`

N/A

7. Please enable verbose logging for your app using `[MSAppCenter setLogLevel: MSLogLevelVerbose];` before your call to `[AppCenterReactNative register];` (or any other SDK registration). For Android, use `AppCenter.setLogLevel(android.util.Log.VERBOSE);` before your `SoLoader.init` call. Include the logs here:

N/A

bmourat:
Fix for #318

### DIFF
--- a/AppCenterReactNativeShared/android/build.gradle
+++ b/AppCenterReactNativeShared/android/build.gradle
@@ -27,7 +27,7 @@ allprojects {
 
     android {
         compileSdkVersion 23
-        buildToolsVersion '23.0.1'
+        buildToolsVersion '27.0.3'
 
         defaultConfig {
             minSdkVersion 16
--- a/appcenter-analytics/android/build.gradle
+++ b/appcenter-analytics/android/build.gradle
@@ -2,15 +2,15 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "27.0.3"
+    buildToolsVersion '27.0.3'
 
     defaultConfig {
         minSdkVersion 16
         targetSdkVersion 22
         versionCode 28
-        versionName "1.5.1"
+        versionName '1.5.1'
         ndk {
-            abiFilters "armeabi-v7a", "x86"
+            abiFilters 'armeabi-v7a', 'x86'
         }
     }
 }
--- a/appcenter-analytics/android/build.gradle
+++ b/appcenter-analytics/android/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    buildToolsVersion "27.0.3"
 
     defaultConfig {
         minSdkVersion 16
--- a/appcenter-crashes/android/build.gradle
+++ b/appcenter-crashes/android/build.gradle
@@ -2,15 +2,15 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "27.0.3"
+    buildToolsVersion '27.0.3'
 
     defaultConfig {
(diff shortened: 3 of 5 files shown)


AppCenterReactNativeShared/android/build.gradle
```diff
@@ -27,7 +27,7 @@ allprojects {
 
     android {
         compileSdkVersion 23
-        buildToolsVersion '23.0.1'
+        buildToolsVersion '27.0.3'
```
guperrot: We must not hardcode a version as it could break in existing projects. It should be managed by linker script dynamically.

appcenter-analytics/android/build.gradle
```diff
@@ -2,7 +2,7 @@ apply plugin: 'com.android.library'
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.1"
+    buildToolsVersion "27.0.3"
```
guperrot: Use single quotes.

dhei:
there is an issue of npm install failing VSTS Build which seems to be same problem as https://github.com/npm/registry/issues/334. Please wait and retry again.

dhei:
@bmourat Looks like https://github.com/facebook/react-native/pull/19257/files is addressing the same issue in React Native and it was already merged. Hopefully they will ship it soon. Though I don't think it will make a huge difference, I am inclined to bump the buildToolsVersion to 26.0.3 rather than 27.x.x. What do you think?

guperrot:
Why using older build if we already tested that it does not impact apps, it's always recommended to have recent build tools that fixes bugs or improve perfs right? @dhei

dhei:
@guperrot Fair enough, I am fine with 27.x.x.

PR #244, opened 2018-02-19
Allow using cocoapods without react native link

ghost:
This helps existing projects which consume dependencies via cocoa pods and not project references. 

I did not change the behavior or react-native link since I'm ok if this is just for people who feel confident to integrate manually.

I'm really stuck on using appcenter since I can't yarn install from git subdirectories, otherwise, I would consume my private branch. Please advise. I'm more than happy to take on a different path.

here's the discussion: https://github.com/Microsoft/AppCenter-SDK-React-Native/issues/239

PR #326, opened 2018-06-10
React/RCTDefines.h not found

mmiani:
I'm having this issue: `React/RCTDefines.h not found` after upgrading CocoaPods to v 1.5.3
According to their [changelog](http://blog.cocoapods.org/CocoaPods-1.5.0/#further-improvements)

> CocoaPods will no longer add all header search paths for each xcconfig generated and instead, it only adds the header search paths for the dependencies of the pod.

So I tried adding React as a dependency in the podspec and that fixed the issue

### DIFF
--- a/appcenter-analytics/ios/appcenter-analytics.podspec
+++ b/appcenter-analytics/ios/appcenter-analytics.podspec
@@ -20,4 +20,5 @@ Pod::Spec.new do |s|
 
   s.dependency 'AppCenterReactNativeShared'
   s.dependency 'AppCenter/Analytics'
+  s.dependency 'React'
 end
--- a/appcenter-crashes/ios/appcenter-crashes.podspec
+++ b/appcenter-crashes/ios/appcenter-crashes.podspec
@@ -20,5 +20,6 @@ Pod::Spec.new do |s|
   s.vendored_frameworks = 'AppCenterReactNativeShared/AppCenterReactNativeShared.framework'
 
   s.dependency 'AppCenterReactNativeShared'
-  s.dependency 'AppCenter/Crashes'  
+  s.dependency 'AppCenter/Crashes'
+  s.dependency 'React'
 end
--- a/appcenter-push/ios/appcenter-push.podspec
+++ b/appcenter-push/ios/appcenter-push.podspec
@@ -20,5 +20,6 @@ Pod::Spec.new do |s|
   s.vendored_frameworks = 'AppCenterReactNativeShared/AppCenterReactNativeShared.framework'
 
   s.dependency 'AppCenterReactNativeShared'
-  s.dependency 'AppCenter/Push'  
+  s.dependency 'AppCenter/Push'
+  s.dependency 'React'
 end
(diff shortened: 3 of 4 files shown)


dhei:
Hey @mmiani,

Thanks for the PR.

I just tried `react-native link` a brand new app created by `react-native init` to link latest version of appcenter sdk with cocoapods v1.5.3 and I don't see any issues. Can you describe a repro step of this issue?

The PR itself looks fine except it's missing a change in *appcenter-analytics/ios/appcenter-push.podspec*. Please make sure you also update that file.

mmiani:
My iOS project is not the standard one created by react-native init. 
My problem is that "react-native link" links the appcenter projects into the project "Libraries folder" but my project consumes dependencies via cocoapods and not project references.
Since your last version, it's possible to use cocoapods without "react-native link" [PR](https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/244) so I've configured all the appcenter packages via cocoapods. (it would be good to include this case in the documentation )

My pod file looks like this:
```
pod 'appcenter', path: '../node_modules/appcenter/ios'
pod 'appcenter-analytics', path: '../node_modules/appcenter-analytics/ios'
pod 'appcenter-crashes', path: '../node_modules/appcenter-crashes/ios'
```
I've added the missing change to appcenter-push, please let me know if you need any further clarification.

dhei:
@mmiani, I struggled to setup a react native project that neither use "react-native link" or "Libraries folder", but only use cocoapods. My problem is when I removed all references in Libraries folder from Xcode, it doesn't know where to find reference. 

Can you share a stripped down version of your RN.iOS app not using "Libraries folder" so that I can reproduce? It could much appreciate to see how to convert a "standard" RN.iOS app create via `react-native init` to your flavor. Thanks.

dhei:
@mmiani, a caveat of not using "react-native link": AppCenter SDKs use prompt during "react-native link" to ask for information:

(1) AppSecret which will be recorded in AppCenter-Config.plist;
(2) `[AppCenterReactNative register]` AppCenter start code in AppDelegate.m.

In other words, without "react-native link", users have to add those code themselves. Basically the last 2 steps from [integrate-the-ios-sdk-manually](https://docs.microsoft.com/en-us/appcenter/sdk/getting-started/react-native#32-ios-only-integrate-the-ios-sdk-manually).

But still, I am interested to see what exactly is required to integrate AppCenter SDKs without "react-native link" even with slightly worse user experience.

mmiani:
Unfortunately, I can't share my iOS project, I'll try to generate a clean iOS project with RN integrated and some dependencies configured via CocoaPods so you can have an idea

oblador:
I'm having the same issue with a fresh app using the CocoaPods integration mentioned here: https://facebook.github.io/react-native/docs/integration-with-existing-apps

dhei:
@oblador, have you tried [Integrate the iOS SDK manually doc](https://docs.microsoft.com/en-us/appcenter/sdk/getting-started/react-native#32-ios-only-integrate-the-ios-sdk-manually) and still having the same issue?

oblador:
@dhei Try with your own `BrownfieldTestApp` project to reproduce the issue.

oblador:
@dhei Did you manage to recreate this using your own example app?

fcerdeiral:
Any updates on this? Having the same issue now with CocoaPods `1.6.0.beta.1`

bmourat:
I've managed to reproduce this, and I can confirm that this PR resolves that issue!

bmourat:
I've managed to reproduce this, and I can confirm that this PR resolves that issue!

PR #341, opened 2018-06-28
Reintroduce firebase

bmourat:


DemoApp/android/app/build.gradle
```diff
@@ -136,6 +136,7 @@ dependencies {
     compile fileTree(dir: "libs", include: ["*.jar"])
     compile 'com.android.support:appcompat-v7:23.0.1'
     compile 'com.facebook.react:react-native:+'  // From node_modules
+    compile 'com.google.firebase:firebase-messaging:10.0.1'
```
MatkovIvan: 1. Why `10.0.1`?
1. What about core module?

DemoApp/android/app/src/main/java/com/demoapp/MainApplication.java
```diff
@@ -53,7 +51,6 @@ public void onCreate() {
     super.onCreate();
 
     AppCenter.setLogLevel(Log.VERBOSE);
-    Push.setSenderId("177539951155");
```
MatkovIvan: You should keep the way to test current approach integration

DemoApp/android/build.gradle
```diff
@@ -6,6 +6,7 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:2.2.3'
+        classpath 'com.google.gms:google-services:3.0.0'
```
MatkovIvan: There is `4.0.1`. Looks like this version doesn't support [new firebase versioning](https://developers.google.com/android/guides/versioning).

DemoApp/appcenter-post-clone.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env bash
 echo "Executing post clone script in `pwd`"
 echo $NPM_RC | base64 --decode > $APPCENTER_SOURCE_DIRECTORY/DemoApp/.npmrc
+echo $GOOGLE_SERVICES_JSON | base64 --decode > $APPCENTER_SOURCE_DIRECTORY/DemoApp/android/app/google-services.json
```
guperrot: We decided to commit google-services.json but removing secrets. You can look at how we did it in Android and use the same format (make sure it contains only the minimum information and empty API keys). That way we can remove the App Center build variable and also not have to worry when we clone the SDK.

guperrot: We decided to commit google-services.json but removing secrets. You can look at how we did it in Android and use the same format (make sure it contains only the minimum information and empty API keys). That way we can remove the App Center build variable and also not have to worry when we clone the SDK.

TestApp/appcenter-post-clone.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
dhei: We don't run AC Build on TestApp, only on DemoApp. No need to add this post clone script to TestApp.

DemoApp/android/app/build.gradle
```diff
@@ -84,7 +84,6 @@ def enableProguardInReleaseBuilds = false
 
 android {
     compileSdkVersion 23
-    buildToolsVersion "23.0.1"
```
dhei: I don't understand why remove `buildToolsVersion`, is it no longer needed other changes in this PR?

DemoApp/android/build.gradle
```diff
@@ -3,16 +3,19 @@
 buildscript {
     repositories {
         jcenter()
+        google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:2.2.3'
```
dhei: Looks like blank app created via `react-native init` still uses v2.2.3. Is there a good reason that we'd better use v3.1.3? The inconvenience of bumping gradle version is Android Studio prompting users to download the newer version, so I am slightly prefer to stay on the default version, but I am open to other opinions.

DemoApp/android/app/build.gradle
MatkovIvan: https://developer.android.com/studio/releases/build-tools

> If you're using Android plugin for Gradle 3.0.0 or higher, your project automatically uses a default version of the build tools that the plugin specifies.

dhei: thanks @MatkovIvan, make sense

TestApp/appcenter-post-clone.sh
guperrot: But we do that on other platforms, at least Android and Xamarin where we test building develop TestApp builds on every push.

guperrot: I guess it's not enough as we need to call `./update-npm-packages` for TestApp. Since it's out of scope I would remove this file for now unless you can make it work quickly on AppCenter build on your branch.

dhei: @guperrot The reason we don't run AC Build is AC Build is on npm@3 which does not recognize our local dependency in the format of `"xxx": "file:xxx-1.6.0.tgz"` (see [link](https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/TestApp/package.json#L11)).

guperrot: @bmourat let's delete the file for now then, thanks.

DemoApp/android/build.gradle
bmourat: This was bumped for compatibility with firebase core and messaging dependencies. And since it is DemoApp which is used mostly for testing purpose, I think we can leave it that way.

dhei: make sense, I am fine with using v3.

DemoApp/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
dhei: Please do not manually edit this file, let Android Studio auto generate. (Because I can tell from the timestamp in the first line is supposed to be auto updated by Android Studio).

TestApp/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-all.zip
```
dhei: same here, do not manually edit this file.

DemoApp/android/gradle/wrapper/gradle-wrapper.properties
dhei: Let's remove the timestamp to avoid confusing because sometime you have to manually edit this file and timestamp get out of sync quickly.

TestApp/android/gradle/wrapper/gradle-wrapper.properties
dhei: Let's remove the timestamp to avoid confusing because sometime you have to manually edit this file and timestamp get out of sync quickly.

guperrot: It will be reinserted next time there is an update but that's ok for now I guess, it does not matter.

PR #350, opened 2018-07-17
Bump to react native v0.56.0

dhei:
- upgrade to react native v0.56.0 in TestApp
- upgrade to gradle v3.1.3 in TestApp and keep gradle v2.3.3 in app center modules

Note: Latest Jest v22.4.4 is not compatible with RN v0.56. A workaround is adopted from https://stackoverflow.com/a/51225530/873234 by adding the following code to package.json:

```
    "transform": {
      "^.+\\.js$": "<rootDir>/node_modules/react-native/jest/preprocessor.js"
    }
```


TestApp/android/build.gradle
```diff
@@ -27,3 +27,15 @@ allprojects {
         }
     }
 }
+
+subprojects {
```
jwallra: Do we still need this workaround with the other fix?

dhei: @jwallra Yes we still need this for external dependency `react-native-fs` and `react-native-image-picker` in TestApp.

`react-native-fs` is on `buildToolsVersion "25.0.0"` and TestApp is on `buildToolsVersion "27.0.3"`, so without this workaround Android Studio will give the following warning:

```
WARNING: The specified Android SDK Build Tools version (25.0.0) is ignored, as it is below the minimum supported version (27.0.3) for Android Gradle Plugin 3.1.3.
```

AppCenterReactNativeShared/android/build.gradle
```diff
@@ -109,8 +111,8 @@ allprojects {
 }
 
 dependencies {
-    compile 'com.android.support:appcompat-v7:23.0.1'
-    compile 'com.microsoft.appcenter:appcenter:1.6.1'
+    implementation 'com.android.support:appcompat-v7:23.0.1'
```
guperrot: Does it work in a blank app?

yeomann:
why we went back? I didn't get it. "Compile is obsolete and is replaced by implementation" would be removed by the end of 2018. package was using latest Gradle and you guys revert it back? This is not true that it should be done for react-native 0.56.0 as I do have current local setup upgraded to Gradle 3+ with "implementation" and SKD 27 all works great on my end with lots of libraries that are linked.

Issue #367, opened 2018-07-25
Linking results in pods being added to wrong target


ds8k:
### **Description**

After linking the appcenter, crashes, and analytics packages, the Podfile lists pods under the wrong target. Our application has an iMessage sticker pack and the pods are being linked to that instead of the main application.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Create a react native application and add a sticker pack target to it
2. Add appcenter packages and link them
3. Look at podfile and see the pods are linked to the wrong target.

Here's a modified version of what our file looks like:
```
# Uncomment the next line to define a global platform for your project
# platform :ios, '9.0'

target 'StickerPack' do
  # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
  # use_frameworks!

  # Pods for StickerPack
  pod 'AppCenter/Crashes', '~> 1.8.0'
  pod 'AppCenter/Analytics', '~> 1.8.0'
  pod 'AppCenterReactNativeShared', '~> 1.7.1'
  
  platform :ios, '9.0'
end

target 'App' do
  # Uncomment the next line if you're using Swift or would like to use dynamic frameworks
  # use_frameworks!
  
  # Pods for App
  
end
```

### **Details**

1. Which SDK version are you using? **1.7.1**
2. Which OS version did you experience the issue on? **N/A**
3. What device version did you see this error on?  Were you using an emulator or a physical device? **N/A**
4. What third party libraries are you using? **N/A**
5. Run the following command and paste the output below: `react-native info`
```
Environment:
  OS: macOS High Sierra 10.13.5
  Node: 8.11.1
  Yarn: 1.7.0
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.4.1 Build version 9F2000
  Android Studio: 3.1 AI-173.4720617

Packages: (wanted => installed)
  react: ^16.3.2 => 16.4.0
  react-native: 0.55.3 => 0.55.3
```

6. If you're developing for React Native iOS, run the following command and paste the output below: `pod --version`
```
$ pod --version
1.5.3
```

bmourat:
Hi @ds8k
Thanks for reporting this, we will take a look!

dhei:
Hey @ds8k,
This issue should be fixed in v1.8.0 (https://github.com/Microsoft/AppCenter-SDK-React-Native/releases/tag/1.8.0). Thank you.

PR #369, opened 2018-07-26
Enhanced pod file integration

bmourat:
Enhance current Pod file integration by adding dependencies to target with the current project name. If such a target is not found, a target is determined the old way. Addresses this issue #367.

### DIFF
--- a/appcenter-link-scripts/src/ios/PodFile.js
+++ b/appcenter-link-scripts/src/ios/PodFile.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const os = require('os');
 const path = require('path');
 const childProcess = require('child_process');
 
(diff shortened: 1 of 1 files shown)


appcenter-link-scripts/src/ios/PodFile.js
```diff
@@ -107,15 +108,25 @@ Podfile.prototype.isTargetPlatformDefined = function (platform) {
...
 
 Podfile.prototype.getTargetSectionPattern = function () {
-    const patterns = this.fileContents.match(/# Pods for .*/);
-    if (patterns === null) {
-        throw new Error(`
-    Error: Could not find a "# Pods for" comment in your Podfile. Please add a "# Pods for AppCenter" line
-    in ${this.file}, inside
-    the "target" section, then rerun the react-native link. AppCenter pods will be added below the comment.
-    `);
+    const projectDirectory = path.resolve(__dirname, '..', '..', '..', '..');
```
dhei: Concat four `'..'` to find the project directory assumes "appcenter-link-scripts" directory is at "/node_modules/appcenter-link-scripts" rather than multiple level down under the "/node_modules" folder like "/node_modules/appcenter/appcenter-link-scripts". This shouldn't be a problem for anyone using npm v3 or later, so it's good. (I tested this pr works with npm v3.10.10, which is the version of npm on App Center Build VMs.)

PR #370, opened 2018-07-30
Remove duplicates from build.gradle when using compile or implementation

bmourat:
A previous version of script removed duplicate dependencies only if they were linked with `compile` keyword. This version enhances this process taking into account dependencies that were linked with `implementation` keyword.

### DIFF
--- a/appcenter-link-scripts/src/android/index.js
+++ b/appcenter-link-scripts/src/android/index.js
@@ -75,7 +75,7 @@ module.exports = {
         const gradlePath = 'android/app/build.gradle';
         let gradleContent = fs.readFileSync(gradlePath, 'utf-8');
         gradleContent.split('\n').forEach((line) => {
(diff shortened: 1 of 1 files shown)


appcenter-link-scripts/src/android/index.js
```diff
@@ -75,11 +75,13 @@ module.exports = {
         const gradlePath = 'android/app/build.gradle';
         let gradleContent = fs.readFileSync(gradlePath, 'utf-8');
         gradleContent.split('\n').forEach((line) => {
-            if (lines[line]) {
-                gradleContent = gradleContent.replace(line, '');
-            }
-            if (line.match(/^\s*compile project.*appcenter.*$/)) {
-                lines[line] = true;
+            let match = line.match(/^\s*(compile|implementation) (project.*appcenter.*)$/);
```
dhei: eslint error "78:17  error  'match' is never reassigned. Use 'const' instead  prefer-const" from VSTS build. Please change `let` to `const`.

PR #377, opened 2018-08-19
Adds api for retrieving transmission targets

bmourat:
Adds api for retrieving transmission targets for Android and iOS
Adds capabilities for testing this API to TestApp

### DIFF
--- a/TestApp/package.json
+++ b/TestApp/package.json
@@ -18,6 +18,7 @@
     "react-native-dialog-component": "^0.2.17",
     "react-native-fs": "^2.9.12",
     "react-native-image-picker": "^0.26.7",
+    "react-native-modal-selector": "0.0.28",
     "react-native-simple-toast": "0.0.8",
     "react-navigation": "^2.11.2"
   },
--- a/appcenter-analytics/Analytics.js
+++ b/appcenter-analytics/Analytics.js
@@ -2,7 +2,7 @@ const ReactNative = require('react-native');
 
 const { AppCenterReactNativeAnalytics } = ReactNative.NativeModules;
 
-module.exports = {
+const Analytics = {
     // async - returns a Promise
     trackEvent(eventName, properties) {
         return AppCenterReactNativeAnalytics.trackEvent(eventName, sanitizeProperties(properties));
@@ -20,28 +20,26 @@ module.exports = {
 
     // async - returns a Promise
     getTransmissionTarget(targetToken) {
-        return new Promise(function(resolve, reject) {
+        return new Promise((resolve, reject) => {
             AppCenterReactNativeAnalytics.getTransmissionTarget(targetToken)
-            .then(targetToken => {
-                resolve(new TransmissionTarget(targetToken));
-            })
-            .catch(error => {
-                reject(error);
-            })
-        })
+            .then(token => resolve(new Analytics.TransmissionTarget(token)))
+            .catch(error => reject(error));
+        });
     },
+};
 
-    TransmissionTarget: class {
-        constructor(targetToken) {
-            this._targetToken = targetToken;
-        }
+Analytics.TransmissionTarget = class {
+    constructor(targetToken) {
+        this.targetToken = targetToken;
+    }
 
-        // async - returns a Promise
-        trackEvent(eventName, properties) {
-            return AppCenterReactNativeAnalytics.trackEventForTransmissionTarget(this._targetToken, eventName, sanitizeProperties(properties));
-        }
+    // async - returns a Promise
+    trackEvent(eventName, properties) {
+        return AppCenterReactNativeAnalytics.trackEventForTransmissionTarget(this.targetToken, eventName, sanitizeProperties(properties));
     }
 };
+
+module.exports = Analytics;
 
 function sanitizeProperties(props = null) {
     // Only string:string mappings are supported currently.
--- a/appcenter-analytics/Analytics.js
+++ b/appcenter-analytics/Analytics.js
@@ -20,15 +20,15 @@ const Analytics = {
 
     // async - returns a Promise
     getTransmissionTarget(targetToken) {
-        return new Promise((resolve, reject) => {
+        return new Promise((resolve) => {
             AppCenterReactNativeAnalytics.getTransmissionTarget(targetToken)
-            .then(token => {
+            .then((token) => {
                 if (!token) {
                     resolve(null);
                 } else {
                     resolve(new Analytics.TransmissionTarget(token));
                 }
-            })
+            });
         });
     },
 };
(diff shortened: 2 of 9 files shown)


TestApp/package.json
```diff
@@ -18,6 +18,7 @@
     "react-native-dialog-component": "^0.2.17",
     "react-native-fs": "^2.9.12",
     "react-native-image-picker": "^0.26.7",
+    "react-native-modal-selector": "0.0.28",
```
MatkovIvan: Is there something wrong with the built-in [Picker](https://facebook.github.io/react-native/docs/picker)?

bmourat: Yes, on ios it can't be collapsed. See [here.](https://stackoverflow.com/questions/41181683/react-native-ios-picker-is-always-open) It's not the bug, but the default's picker behavior.

appcenter-analytics/android/src/main/java/com/microsoft/appcenter/reactnative/analytics/AppCenterReactNativeAnalyticsModule.java
```diff
@@ -60,4 +66,30 @@ public void trackEvent(String eventName, ReadableMap properties, Promise promise
             promise.reject(e);
         }
     }
+
+    @ReactMethod
+    public void trackEventForTransmissionTarget(String targetToken, String eventName, ReadableMap properties, Promise promise) {
```
jwallra: This should just be named `trackEvent`

jwallra: And the `targetToken` parameter should be third to align with other APIs

appcenter-analytics/ios/AppCenterReactNativeAnalytics/AppCenterReactNativeAnalytics.m
```diff
@@ -31,43 +37,86 @@ @implementation AppCenterReactNativeAnalytics
...
-    resolve(nil);
+  [MSAnalytics setEnabled:shouldEnable];
+  resolve(nil);
 }
 
 RCT_EXPORT_METHOD(trackEvent:(NSString *)eventName
-              withProperties:(NSDictionary *)properties
-                    resolver:(RCTPromiseResolveBlock)resolve
-                    rejecter:(RCTPromiseRejectBlock)reject)
+                  withProperties:(NSDictionary *)properties
```
jwallra: Which formatter are you using here?  Our convention (though I don't like it) is to align the `:`s

appcenter-analytics/ios/AppCenterReactNativeAnalytics/AppCenterReactNativeAnalytics.m
```diff
@@ -31,43 +37,86 @@ @implementation AppCenterReactNativeAnalytics
...
+    }
+    return YES;
+  }] allObjects];
+  NSArray * newValues = [properties objectsForKeys:allowedKeys notFoundMarker:@""];
+  NSDictionary * filteredProperties = [NSDictionary dictionaryWithObjects:newValues forKeys:allowedKeys];
+  [MSAnalytics trackEvent:eventName withProperties:filteredProperties];
+  resolve(nil);
+}
+
+RCT_EXPORT_METHOD(trackEventForTransmissionTarget:(NSString *)targetToken
```
jwallra: Align method signature with https://github.com/Microsoft/AppCenter-SDK-Apple/blob/develop/AppCenterAnalytics/AppCenterAnalytics/MSAnalytics.m#L190

appcenter-analytics/Analytics.js
```diff
@@ -16,9 +16,31 @@ module.exports = {
     // async - returns a Promise
     setEnabled(enabled) {
         return AppCenterReactNativeAnalytics.setEnabled(enabled);
+    },
+
+    // async - returns a Promise
+    getTransmissionTarget(targetToken) {
```
guperrot: this should not be async

appcenter-analytics/android/src/main/java/com/microsoft/appcenter/reactnative/analytics/AppCenterReactNativeAnalyticsModule.java
```diff
@@ -16,6 +20,8 @@
 @SuppressWarnings("WeakerAccess")
 public class AppCenterReactNativeAnalyticsModule extends BaseJavaModule {
 
+    private Map<String, AnalyticsTransmissionTarget> transmissionTargets = new HashMap<>();
```
guperrot: mTransmissionTargets

appcenter-analytics/Analytics.js
bmourat: Since all ReactNative bridge interactions are asynchronous, I don't understand how this can not be async?

guperrot: I didn't look like async in the resolve clause you just pass another call, is that async as well? if yes why do we have intermediate code.

guperrot: Forgot about that limitation in React, not ideal but no choice I guess.

appcenter-analytics/android/src/main/java/com/microsoft/appcenter/reactnative/analytics/AppCenterReactNativeAnalyticsModule.java
```diff
@@ -60,4 +66,30 @@ public void trackEvent(String eventName, ReadableMap properties, Promise promise
...
+        } catch (JSONException e) {
+            promise.reject(e);
+        }
+    }
+
+    @ReactMethod
+    public void getTransmissionTarget(String targetToken, Promise promise) {
+        AnalyticsTransmissionTarget transmissionTarget = Analytics.getTransmissionTarget(targetToken);
+        if (transmissionTarget == null) {
+            promise.reject(new IllegalArgumentException("Transmission target may not be null or empty."));
```
guperrot: I know we did inconsistent behaviors in the past but our other native and .NET SDKs are never throwing exceptions to avoid crashing a PROD app just because of a misconfigured telemetry SDK.

For this one native SDK logs an error and return null target and its documented that way.
We should do the same for JS. New JS code should align with native SDKs, we can fix older promises later (never reject anything in new code).

appcenter-analytics/Analytics.js
dhei: +1 for @bmourat, react native is asynchronous. Actually, there is an undocumented way to make blocking synchronous call via [RCT_EXPORT_BLOCKING_SYNCHRONOUS_METHOD](https://github.com/facebook/react-native/blob/master/React/Base/RCTBridgeModule.h#L166-L183), but it's highly discouraged and can be problematic.

PR #378, opened 2018-08-20
Use import instead of inlined require

dhei:


TestApp/app/screens/CrashesScreen.js
```diff
@@ -8,12 +8,12 @@ import Crashes from 'appcenter-crashes';
 import { FooClass } from '../../js/FooClass';
 import AttachmentsProvider from '../AttachmentsProvider';
 import SharedStyles from '../SharedStyles';
+import CrashesTabBarIcon from '../assets/crashes.png';
+
```
MatkovIvan: Extra line

PR #380, opened 2018-08-21
Feature/1ds android config

guperrot:


TestApp/app/screens/AppCenterScreen.js
```diff
@@ -44,6 +59,11 @@ export default class AppCenterScreen extends Component {
     Toast.show('Scheduled custom properties log. Please check verbose logs.');
   }
 
+  async configureStartup(secretString) {
+    await SecretStringHelper.configureStartup(secretString);
+    Toast.show('Startup mode updated. Please kill the application and relaunch again.');
```
jwallra: I would go with something other than "kill".  The wording I used for iOS was "Relaunch application for changes to be applied"

guperrot: @jwallra I just saw your comment and someone merged the PR, I suggest you update your branch based on develop and include the text change in iOS PR.

PR #361, opened 2018-07-23
Fix AppCenterReactNativeShared version in crashes module

dhei:
Mirror #360

PR #385, opened 2018-08-22
Replace GitHub PR numbers with actual links

bmourat:


CHANGELOG.md
```diff
@@ -3,7 +3,7 @@ ___
 ## Version 1.7.1
 
 ### Bugfixes
-- Fix AppCenterReactNativeShared pod version issue for appcenter-crashes module. #361
+- Fix AppCenterReactNativeShared pod version issue for appcenter-crashes module. https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/361
```
MatkovIvan: Let's use `[#361](https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/361)`

jwallra: I agree, let's use the pretty formatting.  I had started doing that in some of the 1DS docs and we should do it here too

PR #386, opened 2018-08-22
Fix SDK to not use reject in SDK methods to not cause crashes

guperrot:


appcenter-analytics/android/src/main/java/com/microsoft/appcenter/reactnative/analytics/AppCenterReactNativeAnalyticsModule.java
```diff
@@ -63,25 +66,23 @@ public void accept(Boolean enabled) {
     public void trackEvent(String eventName, ReadableMap properties, Promise promise) {
         try {
             Analytics.trackEvent(eventName, ReactNativeUtils.convertReadableMapToStringMap(properties));
-            promise.resolve("");
         } catch (JSONException e) {
-            promise.reject(e);
+            AppCenterLog.error(LOG_TAG, "Could not convert event properties from JavaScript to Java", e);
         }
+        promise.resolve(null);
```
dhei: `Analytics.trackEvent(event)` API returns a promise actually. This will change this API to return a resolved promise containing `null` instead of a rejected promise containing the error. It's not quite ideal to get a resolved promise containing `null`. We should think more about it.

dhei: After thinking more about it, since `Analytics.trackEvent(event)` API is supposed to be a void method on native SDKs, it's actually the right thing to return a resolved promise wrapping `null` as result. And native SDKs already logs the error and we don't need to log the same error in JS, so resolving the promise in native bridge code is the right thing to do.

PR #388, opened 2018-08-22
Moving plist info to NSUserDefaults

jwallra:
Plist didn't work on real device so moving info to `NSUserDefaults`


AppCenterReactNativeShared/ios/AppCenterReactNativeShared/AppCenterReactNativeShared.m
```diff
@@ -64,4 +68,8 @@ + (void) setWrapperSdk:(MSWrapperSdk *)sdk {
     [MSAppCenter setWrapperSdk:sdk];
 }
 
++ (void) setStartAutomatically:(BOOL)startedAutomatically {
```
dhei: nit - What is the reason to name it `(BOOL)startedAutomatically`? I prefer `(BOOL)shouldStartAutomatically` or `(BOOL)startAutomatically`.

guperrot: name conflict with the static variable

guperrot:
There is an app freeze when overriding with appsecret only, don't merge yet.

AppCenterReactNativeShared/ios/AppCenterReactNativeShared/AppCenterReactNativeShared.m
guperrot: On Android we have the getter `isStartedAutomatically`, removing the `ed` is odd.
should: then we would have a setter with `setShouldStartAutomatically` which would be odd.

On Android is just the couple is/setStartedAutomatically like every boolean "property".

AppCenterReactNativeShared/ios/AppCenterReactNativeShared/AppCenterReactNativeShared.h
```diff
@@ -14,4 +14,6 @@
 
 + (void) setWrapperSdk:(MSWrapperSdk *)sdk;
 
++ (void) setStartAutomatically:(BOOL)startedAutomatically;
```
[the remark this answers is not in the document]
guperrot: It does not solve the conflict has a static property needs a setter.

AppCenterReactNativeShared/ios/AppCenterReactNativeShared/AppCenterReactNativeShared.m
TroubleMakerBen: Just adding this for completeness:

If we define the static property like this

`@property (class, nonatomic, assign) BOOL isStartedAutomatically;`, provide an ivar `static BOOL _isStartedAutomatically = NO`;
and we define getter and setter like this

```
+ (BOOL)isStartedAutomatically {
  return _isStartedAutomatically;
}

+ (void)setIdentifier:(BOOL)isStartedAutomatically {
  _isStartedAutomatically = isStartedAutomatically;
}

``` there should not be any naming conflicts.

PR #390, opened 2018-08-23
Fix refresh issue in 1st screen

guperrot:
When skip start from app and using library only:
first screen has initial state of disabled
and no installId on Android.
Going to transmission tab starts AppCenter for library.
After that step AppCenter is enabled and
we have access to installId but the first screen was not refreshed.
This change makes sure screen updated when going back to it.
A similar issue was fixed earlier in Java app.

Changed tabs order.

### DIFF
--- a/TestApp/app/screens/AppCenterScreen.js
+++ b/TestApp/app/screens/AppCenterScreen.js
@@ -30,7 +30,7 @@ export default class AppCenterScreen extends Component {
       const refreshAppCenterScreen = navigation.getParam('refreshAppCenterScreen');
 
       // Initial press: the function is not defined yet so nothing to refresh.
-      if (typeof (refreshAppCenterScreen) === 'function') {
+      if (refreshAppCenterScreen) {
         refreshAppCenterScreen();
       }
       defaultHandler();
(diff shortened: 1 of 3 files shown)


TestApp/app/screens/AppCenterScreen.js
```diff
@@ -25,7 +25,16 @@ SecretStrings.android.both = `appsecret=${SecretStrings.android.appSecret};${Sec
 
 export default class AppCenterScreen extends Component {
   static navigationOptions = {
-    tabBarIcon: () => <Image style={{ width: 24, height: 24 }} source={DialsTabBarIcon} />
+    tabBarIcon: () => <Image style={{ width: 24, height: 24 }} source={DialsTabBarIcon} />,
+    tabBarOnPress: ({ defaultHandler, navigation }) => {
+      const refreshAppCenterScreen = navigation.getParam('refreshAppCenterScreen');
+
+      // Initial press: the function is not defined yet so nothing to refresh.
+      if (typeof (refreshAppCenterScreen) === 'function') {
```
dhei: nit - `typeof refreshAppCenterScreen`, not `typeof (refreshAppCenterScreen)`

dhei: `if (refreshAppCenterScreen)` is shorter and works just fine.

guperrot: resolved

PR #393, opened 2018-08-23
Fix local Android release build for TestApp and crash improvements

guperrot:


TestApp/app/screens/CrashesScreen.js
```diff
@@ -1,6 +1,6 @@
 import React, { Component } from 'react';
-import { Image, View, Text, TextInput, Switch, SectionList, TouchableOpacity } from 'react-native';
-import { DialogComponent } from 'react-native-dialog-component';
+import { Image, View, Text, Switch, SectionList, TouchableOpacity } from 'react-native';
+import DialogInput from 'react-native-dialog-input';
```
guperrot: react-native-dialog-component depends on a lib that is broken in release builds when using react native 0.56: https://github.com/magicismight/react-native-root-siblings/issues/29

This lib has the feature we want so replacing.

TestApp/app/App.js
```diff
@@ -84,14 +84,20 @@ Crashes.setListener({
...
         AttachmentsProvider.getBinaryAttachment(),
         AttachmentsProvider.getBinaryName(),
         AttachmentsProvider.getBinaryType(),
       ]);
-      return [ErrorAttachmentLog.attachmentWithText(textAttachment, 'hello.txt'),
-      ErrorAttachmentLog.attachmentWithBinary(binaryAttachment, binaryName, binaryType)];
+      if (textAttachment !== null) {
+        attachments.push(ErrorAttachmentLog.attachmentWithText(textAttachment, 'hello.txt'));
+      }
+      if (binaryAttachment != null && binaryName != null && binaryType != null) {
```
dhei: use `!==` instead of `!=`

TestApp/app/AttachmentsProvider.js
```diff
@@ -45,24 +55,42 @@ export default class AttachmentsProvider {
   }
 
   static async getBinaryAttachmentInfo() {
-    const fileName = await getItemFromStorage(BINARY_FILENAME_KEY);
+    let fileName = await getItemFromStorage(BINARY_FILENAME_KEY);
     const fileSize = await getItemFromStorage(BINARY_FILESIZE_KEY);
-    return `${fileName} (${fileSize})`;
+    if (fileName != null) {
+      if (fileName.length > 16) {
```
dhei: Can you add a comment for this 16? It looks like a magic number.

PR #409, opened 2018-09-03
Add properties UI to TransmissionScreen

bmourat:


TestApp/app/components/PropertiesConfiguratorView.js
```diff
@@ -0,0 +1,50 @@
...
+      <View style={{ backgroundColor: 'white', padding: 10, flexDirection: 'column' }}>
+        <TouchableOpacity onPress={() => { if (this.props.allowChanges) this.props.onPropertyAdded(); }}>
+          <View style={{ flexDirection: 'row', alignItems: 'center' }}>
+            <Image style={{ height: 24, width: 24 }} source={AddIcon} />
+            <Divider />
+            <Text>Add property</Text>
+          </View>
+        </TouchableOpacity>
+        <FlatList
+          style={{ marginTop: 10 }}
```
MatkovIvan: Why do you use some styles from the `StyleSheet` but some ones just inline? Let's use it uniformly.

TestApp/app/screens/TransmissionScreen.js
```diff
@@ -41,6 +81,19 @@ export default class TransmissionScreen extends Component {
         <Text style={SharedStyles.itemButton}>{title}</Text>
       </TouchableOpacity>
     );
+    const propertiesRenderItem = () => (
+      <PropertiesConfiguratorView
+        onPropertyAdded={() => {
+        const nextItem = this.state.properties.length + 1;
+        this.addProperty({ name: `key${nextItem}`, value: `value${nextItem}` });
+}
```
MatkovIvan: Please format

TestApp/app/screens/TransmissionScreen.js
```diff
@@ -55,7 +108,13 @@ export default class TransmissionScreen extends Component {
               data: [
                 {
                   title: this.state.targetToken.label,
-                  valueChanged: option => this.setState({ targetToken: option }),
+                  valueChanged: (option) => {
+                    this.setState({
+                      targetToken: option,
+                      showProperties: !!option.key,
+                      properties: this.targetTokensProperties[option.key]
+});
```
MatkovIvan: Please format

bmourat:
For some reason, if this line 
```
style={{ marginTop: 10 }}
```
is imported from styles, app state is not refreshed after `add property` button click

TestApp/app/screens/TransmissionScreen.js
```diff
@@ -18,14 +20,53 @@ export default class TransmissionScreen extends Component {
...
+    await target.propertyConfigurator.setEventProperty(property.name, property.value);
+    this.setState((state) => {
+      state.properties.push(property);
+      this.targetTokensProperties[this.state.targetToken.key] = state.properties;
+      return state;
+    });
+  }
+
+  async removeProperty(propertyName) {
+    const target = await Analytics.getTransmissionTarget(this.state.targetToken.key);
```
guperrot: Is inheritance coming in a separate PR?

Also it uses the new APIs but I see the PR targets develop and not `feature/transmission-target-properties-ui`, how did you test and could lint pass by calling non existing APIs?

bmourat: We've developed and tested this with API in #412, but divided in 2 PR. So, I think this should be merged only after #412 is merged.

guperrot: What about inheritance? Is this coming after? Today the app gets all targets at root level.

TestApp/app/screens/TransmissionScreen.js
```diff
@@ -18,14 +20,53 @@ export default class TransmissionScreen extends Component {
...
+    await target.propertyConfigurator.removeEventProperty(propertyName);
+    this.setState((state) => {
+      state.properties = state.properties.filter(item => item.name !== propertyName);
+      this.targetTokensProperties[this.state.targetToken.key] = state.properties;
+      return state;
+    });
+  }
+
+  async replaceProperty(oldPropertyName, newProperty) {
+    const target = await Analytics.getTransmissionTarget(this.state.targetToken.key);
```
guperrot: Implement inheritance like specified in the UI instead of always using root.

guperrot: We can remove null/empty targets from the UI to simplify.

Issue #407, opened 2018-09-01
1.8.0 - AppCenter not enabled on start resulting in null install ID


CodyBrouwers:
### **Description**

After upgrading to version 1.8.0 of the SDK, calling `AppCenter.getInstallId()` results in a `null` value in Android because AppCenter is not enabled on start. I believe it has to do with PR https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/383 that looks like it implements some sort of auto start option. I wasn't able to get AppCenter to be enabled on app load unless I downgraded to 1.7.1

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Upgrade to 1.8.0
2. Start Android app, simulator or device.
3. Check if AppCenter is started using `isEnabled()`, I get false
4. Since it isn't started, `getInstallId()` returns `null`

### **Details**

1. Which SDK version are you using?
    - 1.8.0
2. Which OS version did you experience the issue on?
    - Android 8.1.0
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - Google Pixel 2 and Android Emulator

<details><summary>5. `react-native info` command output:</summary>
<p>

```
  React Native Environment Info:
    System:
      OS: macOS High Sierra 10.13.6
      CPU: x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
      Memory: 764.03 MB / 16.00 GB
      Shell: 2.7.1 - /usr/local/bin/fish
    Binaries:
      Node: 8.11.3 - ~/.nvm/versions/node/v8.11.3/bin/node
      Yarn: 1.9.4 - ~/.nvm/versions/node/v8.11.3/bin/yarn
      npm: 6.1.0 - ~/.nvm/versions/node/v8.11.3/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 11.4, macOS 10.13, tvOS 11.4, watchOS 4.3
      Android SDK:
        Build Tools: 23.0.1, 25.0.0, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.1, 27.0.2, 27.0.3
        API Levels: 23, 24, 25, 26, 27
    IDEs:
      Android Studio: 3.0 AI-171.4443003
      Xcode: 9.4.1/9F2000 - /usr/bin/xcodebuild
    npmPackages:
[...]

PR #410, opened 2018-09-03
Fix regression in Java shared module config

guperrot:
Fix for bug #407 and some refactoring.

### DIFF
--- a/AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
+++ b/AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
@@ -15,67 +15,51 @@ import static com.microsoft.appcenter.utils.AppCenterLog.LOG_TAG;
 
 public class AppCenterReactNativeShared {
 
+    private static final String APPCENTER_CONFIG_ASSET = "appcenter-config.json";
+
     private static final String APP_SECRET_KEY = "app_secret";
 
     private static final String START_AUTOMATICALLY_KEY = "start_automatically";
+
+    private static Application sApplication;
 
     private static String sAppSecret;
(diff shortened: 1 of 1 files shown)


AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
```diff
@@ -15,67 +15,51 @@
...
+        wrapperSdk.setWrapperSdkName(com.microsoft.appcenter.reactnative.shared.BuildConfig.SDK_NAME);
+        AppCenter.setWrapperSdk(wrapperSdk);
+        readConfigurationFile();
+        if (!sStartAutomatically) {
+            AppCenterLog.debug(LOG_TAG, "Configure not to start automatically.");
             return;
         }
 
         /* Get app secret from appcenter-config.json file. */
-        String appSecret = getAppSecret();
```
guperrot: That's where the bug was, we reading the file after reading `sStartAutomatically`, chicken/egg issue. iOS code was doing that in the right order.

AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
```diff
@@ -84,16 +68,19 @@ public static String getAppSecret() {
                 configStream.close();
                 String jsonContents = new String(buffer, "UTF-8");
                 JSONObject json = new JSONObject(jsonContents);
-                sAppSecret = json.getString(APP_SECRET_KEY);
+                sAppSecret = json.optString(APP_SECRET_KEY);
```
guperrot: To align with iOS where missing key in plist works, optString returns empty if key missing

PR #419, opened 2018-09-10
Add UI for device id to test application

MatkovIvan:


guperrot:
We cherry-picked db0f306 to include it in #421.
The other changes are too big and not needed.

PR #446, opened 2018-10-18
Remove package-lock.json

bmourat:


dhei:
FYI, package-lock.json is removed due to unreliable build caused by package-lock.json. We don't know exactly what caused the below npm error with `npm install`, but removing package-lock.json resolved the issue.

```
npm ERR! Maximum call stack size exceeded
```

Issue #451, opened 2018-10-25
gradlew check error


yeomann:
After creating build, my app is keep crashing. Not sure who is the correct culprit although gradlew check is giving me hints from App center. help me for possible fix.
my ext looks like this
```
ext {
    buildToolsVersion = "28.0.3"
    minSdkVersion = 19
    compileSdkVersion = 27
    targetSdkVersion = 27
    supportLibVersion = "28.0.0"
    googlePlayServicesVersion = "15.0.1"
}
```

### **Description**
Hellow `./gradlew check` is failing
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':appcenter:lint'.

node_modules/appcenter/android/build.gradle:9: Error: Google Play will soon require that apps target API level 26 or higher. This will be required for new apps in August 2018, and for updates to existing apps in November 2018. [ExpiringTargetSdkVersion]
          targetSdkVersion 22
          ~~~~~~~~~~~~~~~~~~~
```

### **Details**

Which version of the App Center SDK are you using?
```
    "appcenter": "^1.7.1",
    "appcenter-analytics": "^1.7.1",
    "appcenter-crashes": "^1.7.1",
```

React Native Info
```
  React Native Environment Info:
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i7-3720QM CPU @ 2.60GHz
      Memory: 105.00 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 9.11.1 - /usr/local/bin/node
      Yarn: 1.9.4 - /usr/local/bin/yarn
      npm: 6.4.1 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 5.0
      Android SDK:
        Build Tools: 23.0.1, 25.0.1, 25.0.2, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.1, 27.0.2, 27.0.3, 28.0.0, 28.0.3
        API Levels: 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28
    IDEs:
      Android Studio: 3.1 AI-173.4907809
      Xcode: 10.0/10A255 - /usr/bin/xcodebuild
    npmPackages:
[...]

Issue #365, opened 2018-07-24
Firebase-messaging reintroduced with old version


aaroncmoore:
I ran into this issue tonight while building my react native app with firebase:

Failed to capture snapshot of input files for task ':appcenter-push:compileReleaseAidl' property 'importDirs' during up-to-date check.
> The library com.google.firebase:firebase-iid is being requested by various other libraries at [[16.2.0,16.2.0]], but resolves to 16.0.0. Disable the plugin and check your dependencies tree using ./gradlew :app:dependencies.

Confused what or how appcenter-push was creating this issue I did some digging for updates and found this:

https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/341/files/6be2c0d454a33dc39e560538ebd445a23d0d43d0

firebase-messaging was re-introduced into appcenter with:
+    implementation 'com.google.firebase:firebase-core:16.0.1'
+    implementation 'com.google.firebase:firebase-messaging:17.0.0'

The version of messaging is not up to date (17.1.0 is the up to date version).  Firebase messaging 17.0.0 uses firebase-iid 16.0.0, while firebase 17.1.0 uses firebase-iid 16.2.0, thus creating the conflict.

Would it be possible to use firebase-messaging 17.1.0?  Is there a reason the new version was released with an older firebase version?

The steps to re-produce this is easy enough.  Create a project using 17.1.0 and then compile with appcenter 1.7 and this conflict should appear.

For now I have downgraded to appcenter 1.6 to resolve this issue for my builds but I'd prefer to not have to do this long term.

Thanks!

dhei:
Hey @aaroncmoore,

Thanks for getting in touch. We found the issue that is caused by Google Firebase library dependencies.

The issue is the 'com.google.firebase:firebase-core:16.0.1' takes a dependency on 'com.google.firebase:firebase-iid:16.0.0' while 'com.google.firebase:firebase-messaging:17.1.0' takes a dependency on 'com.google.firebase:firebase-iid:16.2.0'. Starting from 'com.google.gms:google-services:3.3.0', Google is requiring [strict version matching](https://developers.google.com/android/guides/versioning#google_services_plugin). For more information, you can check out the [May 2nd 2018 release notes of Google Services Gradle Plugin](https://developers.google.com/android/guides/releases#may_2_2018) explaining this change.

[...]

PR #452, opened 2018-10-27
Avoid version conflict and lint issues in host application

guperrot:

## Description

* Use host app versions for various android settings when possible to avoid lint errors.
* Also use latest version of Firebase and let react native pull support lib to avoid conflicts.
* Verified lint and runtime of TestApp.
* Verified lint and runtime of new blank app using RN 0.57.4
* Verified publishing 1.9.0-0 temp version to bintray snapshot with the change for shared android module and consumed that version from TestApp.

## Related PRs or issues

#451 and #365

## Misc

Not sure about the impact of older versions of RN yet.


guperrot:
The push part of the fix works only for recent react native versions. Need to find another solution for push gradle.

guperrot:
We could document to add recent support libraries explicitly for old react native versions.

guperrot:
Could not fully automate all lint fixes but now push works without missing dependencies and without lint errors.

To fix lint errors at `:app:lint` level caused by Firebase VS React Native, the application needs to manually add the dependencies of support libraries that mismatch versions. Only customers that want to resolve lint issue will need extra steps in `app/build.gradle`.

Still need testing on old RN versions.

AppCenterReactNativeShared/android/build.gradle
```diff
@@ -109,7 +111,6 @@ allprojects {
 }
 
 dependencies {
-    compile 'com.android.support:appcompat-v7:23.0.1'
```
dhei: 👍 I like removing this -- it shouldn't be there in the first place.

AppCenterReactNativeShared/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Tue Jul 17 18:39:49 PDT 2018
```
dhei: nit-picking, use Android Studio to auto upgrade this.

dhei: ok, never mind

guperrot:
It still works at runtime when using this patch on RN 0.34 / 0.50. Lint will work only on latest RN version though (0.57.4 as of now).

PR #461, opened 2018-11-15
App Center Analytics unit testing

bmourat:

## Description

This PR adds unit testing setup and some unit tests to the analytics module.


appcenter-analytics/package.json
```diff
@@ -36,9 +36,20 @@
     "eslint-config-airbnb": "^15.1.0",
     "eslint-plugin-import": "^2.7.0",
     "eslint-plugin-jsx-a11y": "^6.0.2",
-    "eslint-plugin-react": "^7.2.1"
+    "eslint-plugin-react": "^7.2.1",
+    "jest": "^23.6.0",
+    "react": "^16.6.3",
```
guperrot: This will update react native for the application because of the SDK? This will have side effects on applications using older versions of React Native, and React Native introduces breaking changes.

bmourat: It is a dev dependency, it won't have any effect for users of the sdk

dhei: Like @bmourat said, dev dependency shouldn't be a problem. In case sdk users don't want to conflict RN version, they can use `npm install --production` to skip installing dev dependencies.

guperrot: Most people probably don't know about `--production`, please test on an app that uses old react native version like TestApp34 and see what happens without using that flag.

appcenter-analytics/__tests__/Analytics.js
```diff
@@ -0,0 +1,138 @@
...
+    const spy = jest.spyOn(NativeModules.AppCenterReactNativeAnalytics, 'trackEvent');
+    await Analytics.trackEvent(eventName);
+    expect(spy).toHaveBeenCalledWith(eventName, {});
+  });
+
+  test('trackEvent with invalid properties throws error', async () => {
+    const eventName = 'Test event';
+    const eventProperties = { invalid: { prop: 1 } };
+    expect(() => Analytics.trackEvent(eventName, eventProperties))
+    .toThrowError('Properties cannot be serialized. Object must only contain strings');
```
jwallra: This line should be tabbed over

bmourat:
Tested changes with TestApp34 and can confirm that there are no side effects. PR that updates TestApp34 #470

Issue #472, opened 2018-11-24
AppCenter LogLevel typescript type definition problem


thomasgazzoni:
### **Description**

The Type Definition for the Log Level in the AppCenter package seems wrong.

### **Repro Steps**
If we follow the AppCenter type definition we should use the AppCenterLogLevel enum in this way:

```ts
import AppCenter from 'appcenter';

AppCenter.setLogLevel(AppCenter.AppCenterLogLevel.DEBUG);
```

But in reality the AppCenterLogLevel in undefined because is not defined in the AppCenter.js, so we should use LogLevelXXX instead:

```ts
import AppCenter from 'appcenter';

AppCenter.setLogLevel(AppCenter.LogLevelDebug);
```

But this will rise a type error since Property 'LogLevelDebug' does not exist on type AppCenter.

I think a possible solution could be export all the log level of the appcenter-log.js as a constant in the AppCenter.d.ts type definition file instead of a enum, something like this:

```ts
/**
 * Debug information will be logged
 */
export const LogLevelDebug = 3;
```

### **Details**

AppCenter Package Version:

├─ appcenter-analytics@1.10.0
│  └─ appcenter@1.10.0
├─ appcenter-crashes@1.10.0
│  └─ appcenter@1.10.0
├─ appcenter-link-scripts@1.10.0
├─ appcenter@1.10.0
│  └─ appcenter-link-scripts@1.10.0

React Native Environment Info:
    System:
      OS: macOS 10.14
      CPU: x64 Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
      Memory: 546.52 MB / 16.00 GB
      Shell: 5.3 - /bin/zsh
    Binaries:
      Node: 10.5.0 - /usr/local/bin/node
      Yarn: 1.3.2 - ~/.npm-global/bin/yarn
      npm: 6.1.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        Build Tools: 23.0.1, 23.0.2, 23.0.3, 25.0.0, 25.0.2, 25.0.3, 26.0.1, 26.0.2, 26.0.3, 27.0.0, 27.0.1, 27.0.3, 28.0.3
        API Levels: 23, 24, 25, 26, 27, 28
    IDEs:
[...]

PR #475, opened 2018-11-28
Deprecate redundantly named AppCenterLogLevel in favor of LogLevel

jwallra:

## Description

Previously to access log levels in react native you had to use `AppCenter.AppCenterLogLevel.*`.  We've shortened this by adding an additional identical type `LogLevel` and marking the previous one as deprecated.  `AppCenterLogLevel` will be maintained for backwards compatibility.

## Related PRs or issues

#472


TestAppTypescript/package.json
```diff
@@ -12,11 +12,11 @@
...
-    "appcenter-crashes": "file:appcenter-crashes-1.8.1.tgz",
-    "appcenter-link-scripts": "file:appcenter-link-scripts-1.8.1.tgz",
-    "appcenter-push": "file:appcenter-push-1.8.1.tgz",
+    "appcenter": "file:appcenter-1.10.0.tgz",
+    "appcenter-analytics": "file:appcenter-analytics-1.10.0.tgz",
+    "appcenter-crashes": "file:appcenter-crashes-1.10.0.tgz",
+    "appcenter-link-scripts": "file:appcenter-link-scripts-1.10.0.tgz",
+    "appcenter-push": "file:appcenter-push-1.10.0.tgz",
     "react": "16.3.0-alpha.1",
     "react-native": "0.54.4",
```
dhei: Not absolutely required but would be nice to update `react` and `react-native` to the same version in TestApp (https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/TestApp/package.json#L16-L17)

```
"react": "16.6.0-alpha.8af6728",
"react-native": "0.57.3",
```

guperrot: Can we do that later? Would require to test everything again @dhei

dhei: Fair enough, let's do it later.

Issue #287, opened 2018-04-20
Support delay push notification permission dialog


oskaratcodedesign:
We want to delay AppCenterReactNativePush register, so notification permission is not asked at startup. But it doesnt work:

if Push.setListener({onPushNotificationReceived: pushNotification => {
is called before AppCenterReactNativePush register is finished =
Crash `AppCenterPushNotificationReceived` is not a supported event type for AppCenterReactNativePush. Supported events are: `(null)`

AppCenterReactNativePush needs to be initialized?

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Dont call [AppCenterReactNativePush register]
2. Call Push.setListener({onPushNotificationReceived: pushNotification => { })
or
1. Call [AppCenterReactNativePush register] (eg make a bridge and call it form react js)
and call Push.setListener({onPushNotificationReceived: pushNotification => { })
= either crash or onPushNotificationReceived is never received.
or 
1 If you delay the Push.setListener({onPushNotificationReceived: pushNotification => { }), it doesnt crash, but then onPushNotificationReceived is never received?

### **Details**

1. Which SDK version are you using?
    - appcenter-push@^1.4.0:
2. Which OS version did you experience the issue on?
    - iOS
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - iPhone X physical device
4. What third party libraries are you using?
    - 
5. Run the following command and paste the output below: `react-native info`

Environment:
  OS: macOS High Sierra 10.13.4
  Node: 9.3.0
  Yarn: 1.3.2
  npm: 5.6.0
  Watchman: 4.9.0
  Xcode: Xcode 9.3 Build version 9E145
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: ^16.3.0-alpha.1 => 16.3.0-alpha.1
  react-native: 0.54.2 => 0.54.2

6. If you're developing for React Native iOS, run the following command and paste the output below: `pod --version`
1.5.0

dhei:
Hi @oskaratcodedesign,

[...]

PR #480, opened 2018-12-08
Add configuration to delay the first start of Push until Push.setEnabled(true) is called from JS

guperrot:

## Description

To delay start push with this change:

* Add EnablePushInJavascript YES to AppCenter-Config.plist
* Add `"enable_push_in_javascript": true` to appcenter-config.json
* Call `Push.setEnabled(true)` from Javascript when developer wants to start Push.

Since this is only for the first time we want to delay Push, the code also remembers the first call to `Push.setEnabled(true)` and will start Push automatically the next time.

Documentation PR: https://github.com/MicrosoftDocs/appcenter-docs-pr/pull/1000

Requested in #287


CHANGELOG.md
```diff
@@ -1,11 +1,15 @@
...
-## Version 1.10.1 (under active development)
+## Version 1.11.0 (under active development)
 
 ### AppCenter
 
 - Introduce new `LogLevel` constants, deprecating old ones.
 
+### AppCenterPush
+
+- **[Feature]** Add manual configuration to delay the first start of Push module.
```
dhei: I suggest something like "Support delaying Push notification permission dialog". The original feature request is at https://github.com/Microsoft/AppCenter-SDK-React-Native/issues/287.

appcenter-analytics/android/build.gradle
```diff
@@ -21,6 +21,6 @@ dependencies {
     compile 'com.facebook.react:react-native:+'
     compile 'com.microsoft.appcenter:appcenter-analytics:1.10.0'
 
-    //compile project(':AppCenterReactNativeShared')   // For testing with TestApp
-    compile 'com.microsoft.appcenter.reactnative:appcenter-react-native:1.10.0'
+    compile project(':AppCenterReactNativeShared')   // For testing with TestApp
```
dhei: revert this change.

appcenter-push/android/src/main/java/com/microsoft/appcenter/reactnative/push/AppCenterReactNativePushModule.java
```diff
@@ -10,18 +10,39 @@
...
         AppCenterReactNativeShared.configureAppCenter(application);
         if (AppCenter.isConfigured()) {
-            AppCenter.start(Push.class);
+            boolean startPush = true;
+            boolean enablePushInJavascript = AppCenterReactNativeShared.getConfiguration().optBoolean(ENABLE_PUSH_IN_JAVASCRIPT);
+            if (enablePushInJavascript) {
+
+                /*
+                 * TODO expose a way to post command in background looper in native SDK to avoid accessing storage directly here.
+                 * Storage might not be ready yet.
```
jwallra: What happens if storage is not ready?  Would we crash, or would the push configuration just not be set?

appcenter-push/android/src/main/java/com/microsoft/appcenter/reactnative/push/AppCenterReactNativePushModule.java
```diff
@@ -37,11 +58,20 @@ public String getName() {
...
+            AppCenter.start(Push.class);
+            mPushStarted = true;
+        }
         Push.setEnabled(enabled).thenAccept(new AppCenterConsumer<Void>() {
 
             @Override
             public void accept(Void result) {
+                if (enabled) {
+
+                    /* TODO expose a way to post command in background looper in native SDK to avoid accessing storage directly here. */
```
jwallra: Does this have the same condition ("storage might not be ready yet") as above?  If so, let's include that in the comment.

appcenter-analytics/android/build.gradle
guperrot: Won't be able to compile if I do that, we need to keep that until release.

appcenter-push/android/src/main/java/com/microsoft/appcenter/reactnative/push/AppCenterReactNativePushModule.java
guperrot: We make the storage ready in the next line.

appcenter-push/android/src/main/java/com/microsoft/appcenter/reactnative/push/AppCenterReactNativePushModule.java
guperrot: @jwallra no as storage is initialized in the constructor of this class.
@achocron it should not be a blocker for this work (we already read a file in shared directly = disk access so would require more changes to the RN SDK overall to comply with strict mode) item but yes we can add a work item for that.

appcenter-analytics/android/build.gradle
guperrot: Your feedback is addressed in https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/480/commits/a3adae58e3a62b77aa2185e7ff974beb9a24c3db

scripts/bump-android-sdk-version.sh
```diff
@@ -35,19 +35,14 @@ echo "React-Native Android version $oldAndroidSdkVersion will be updated to $new
...
-
-gradleFileContent="$(cat ./appcenter-analytics/android/build.gradle)"
-echo "${gradleFileContent/com.microsoft.appcenter\:appcenter-analytics\:$oldAndroidSdkVersion/com.microsoft.appcenter:appcenter-analytics:$newAndroidSdkVersion}" > ./appcenter-analytics/android/build.gradle
-
-gradleFileContent="$(cat ./appcenter-push/android/build.gradle)"
-echo "${gradleFileContent/com.microsoft.appcenter\:appcenter-push\:$oldAndroidSdkVersion/com.microsoft.appcenter:appcenter-push:$newAndroidSdkVersion}" > ./appcenter-push/android/build.gradle
-
-gradleFileContent="$(cat ./AppCenterReactNativeShared/android/build.gradle)"
-echo "${gradleFileContent/com.microsoft.appcenter\:appcenter\:$oldAndroidSdkVersion/com.microsoft.appcenter:appcenter:$newAndroidSdkVersion}" > ./AppCenterReactNativeShared/android/build.gradle
+for file in \
```
dhei: I like this improvement in bump-android-sdk-version.sh and bump-wrapper-sdk-version.sh. We should do the same for bump-ios-sdk-version.sh.

jdnichollsc:
`EnablePushInJavascript` is not working for iOS, the event **onPushNotificationReceived** is not executed in that platform

guperrot:
@jdnichollsc it works for us in TestApp from this repo but it looks like your issue is caused by something else since you commented on #505.

jdnichollsc:
@guperrot thanks, I'm reviewing the code of AppCenter but I can't see anything

guperrot:
We will investigate and track in #505 even if not related (can still rename the title there when we find the cause).

Also you mentioned in #505 that version 1.8.0 of the SDK also has the issue so that means it's not caused by this new configuration feature and is a older issue.

PR #493, opened 2018-12-17
Fix initial state of transmission target in demo apps

MatkovIvan:

### Repro Steps

- Install SDK test React Native Android 1.0(215) app
- Launch the app and click Transmission tab
- Disabled Transmission Target "Parent Target Token" 
- Kill the app and restart it 
- Re-click Transmission tab to observe the "Parent Target Token" status
- Switch to Child or Grandchild Target to observe the status
- Switch back to the Parent Target to observe the status

### Expected Behaviour
Parent Transmission Target should be disabled when changed it to disabled and restart it.

### Actual Behaviour
Parent Transmission Target displayed enabled status after switching to disabled and restart the app.


guperrot:
Merged by #497

PR #504, opened 2019-01-14
Add userId support

Stephanie-Dover:

* [X] Has `CHANGELOG.md` been updated?
* [X] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

AppCenter SDK now supports a 256 character User ID string that applies to crashes, error, and push logs. Settable via AppCenter.setUserId(). This feature adds an API, but is not yet supported on the App Center backend.

## Related PRs or issues

List related PRs and other issues.

## Misc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
     * Update react-native to version 0.58+ and update App Center SDK packages.
     * Edit **app/build.gradle** to replace `compile` by `implementation`.
     * Run `react-native link` again.
-- **[Feature]** Allow users to set userId that applies to crashes, error and push logs. This feature adds an API, but is not yet supported on the App Center backend.
+- **[Feature]** Allow users to set userId that applies to crashes, error and push logs.
 
 ## Version 1.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,5 @@
 # App Center SDK for React Native Change Log
 
-### AppCenter
-
 ## Version 1.11.1
 
 ### AppCenter
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # App Center SDK for React Native Change Log
 
+### AppCenter
+
 ## Version 1.11.1
 
 ### AppCenter
@@ -9,6 +11,16 @@
     * Update react-native to version 0.58+ and update App Center SDK packages.
     * Edit **app/build.gradle** to replace `compile` by `implementation`.
     * Run `react-native link` again.
+
+#### Android
+**[Feature]** AppCenter SDK now supports a 256 character User ID string that applies to crashes, error, and push logs. 
+Settable via `AppCenter.SetUserId()`. 
+This feature adds an API, but is not yet supported on the App Center backend.
+
+#### iOS
+**[Feature]** AppCenter SDK now supports a 256 character User ID string that applies to crashes, error, and push logs. 
+Settable via `AppCenter.SetUserId()`. 
+This feature adds an API, but is not yet supported on the App Center backend.
 
 ## Version 1.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,16 +11,7 @@
     * Update react-native to version 0.58+ and update App Center SDK packages.
     * Edit **app/build.gradle** to replace `compile` by `implementation`.
     * Run `react-native link` again.
-
-#### Android
-**[Feature]** AppCenter SDK now supports a 256 character User ID string that applies to crashes, error, and push logs. 
-Settable via `AppCenter.SetUserId()`. 
-This feature adds an API, but is not yet supported on the App Center backend.
-
-#### iOS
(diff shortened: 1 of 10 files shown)


appcenter/AppCenter.d.ts
```diff
@@ -31,5 +31,6 @@ export function getLogLevel(): Promise<LogLevel | AppCenterLogLevel>;
 export function getSdkVersion(): string;
 export function isEnabled(): Promise<boolean>;
 export function setEnabled(enabled: boolean): Promise<void>;
+export function setUserId(string: boolean): Promise<void>;
```
dhei: `setUserId(string: boolean)` should be `setUserId(userId: string)`

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # App Center SDK for React Native Change Log
 
+### AppCenter
```
DennisPan: What's this for?

CHANGELOG.md
```diff
@@ -9,6 +11,7 @@
     * Update react-native to version 0.58+ and update App Center SDK packages.
     * Edit **app/build.gradle** to replace `compile` by `implementation`.
     * Run `react-native link` again.
+- Support a 256 character User ID string that applies to crashes, error, and push logs. Settable via `AppCenter.setUserId()`.
```
DennisPan: >a 256 character User ID string

This could be misleading. When I read it, I read it as the user string must be 256 characters. We should describe it so that it is clear 256 characters is maximum

appcenter/AppCenter.d.ts
youngbupark: Fixed.

TestApp/app/screens/AppCenterScreen.js
```diff
@@ -225,9 +225,17 @@ export default class AppCenterScreen extends Component {
               data: [
                 { title: 'Install ID', value: 'installId' },
                 { title: 'SDK Version', value: 'sdkVersion' },
+                {
+                  title: 'User ID',
+                  value: 'userId',
+                  onChange: async (userId) => {
```
DennisPan: Why there are two `onChange`?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
...
```
dhei: Reuse the text from [Apple SDK changelog](https://github.com/Microsoft/AppCenter-SDK-Apple/blob/develop/CHANGELOG.md#appcenter-1):
> [Feature] Allow users to set userId that applies to crashes, error and push logs. This feature adds an API, but is not yet supported on the App Center backend.

youngbupark: @dhei Thanks for pointing it out.

CHANGELOG.md
youngbupark: Fixed it to what @dhei recommended.

CHANGELOG.md
youngbupark: Fixed the comment.

TestApp/app/screens/AppCenterScreen.js
youngbupark: won't fix.

dhei: It's not duplicated. https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/504/files#diff-9be630c1ef08339c52b5f05a2fa83720R149 is inlined function definition and https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/504/files#diff-9be630c1ef08339c52b5f05a2fa83720R231 is the parameters passed into the function.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
     * Update react-native to version 0.58+ and update App Center SDK packages.
     * Edit **app/build.gradle** to replace `compile` by `implementation`.
     * Run `react-native link` again.
+- **[Feature]** Allow users to set userId that applies to crashes, error and push logs. This feature adds an API, but is not yet supported on the App Center backend.
```
nevalenny: nit: `error` -> `errors` to keep the enumeration plural.

nevalenny: I'm not sure it's a good idea to keep this statement in the version release notes: `This feature adds an API, but is not yet supported on the App Center backend.`

We're going to support User Id in backend soon, maybe even before some users will see this SDK version. I'm not sure if we can retroactively change the version notes later or what is our plan here?

youngbupark: > nit: `error` -> `errors` to keep the enumeration plural.

looks fine grammatically because (logs) is moved out e.g. error **(logs)** and push logs

When it comes to the backend comment, it was @dhei's suggestion to use native change log statement. let's wait for PM's comment.

Stephanie-Dover: fixed by removing `This feature adds an API, but is not yet supported on the App Center backend.`

dhei: @nevalenny In general we don't retroactively update CHANGELOG. CHANGELOG only reflects the changes in a specific sdk version.
It really depends on the timeline of sdk release and backend support. If sdk is released before backend support, let's keep the statement to avoid confusion. Otherwise, let's get rid of "This feature adds an API, but is not yet supported on the App Center backend".

Issue #505, opened 2019-01-14
onPushNotificationReceived not firing for iOS


jdnichollsc:
### **Description**

We're receiving push notifications in our react native app, in Android works perfect (also using a redirection) but from iOS the `onPushNotificationReceived` is not executed (background or foreground).

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Put the configuration:
```
Push.setListener({
    onPushNotificationReceived(pushNotification) {
    ...
    if (AppState.currentState === 'active') {
      Alert.alert(title, message);
    } else {
      // Sometimes the push callback is received shortly before the app is fully active in the foreground.
      // In this case you'll want to save off the notification info and wait until the app is fully shown
      // in the foreground before displaying any UI. You could use AppState.addEventListener to be notified
      // when the app is fully in the foreground.

      // Showing an alert when not in the "active" state seems to work on iOS; for Android, we show a toast
      // message instead
      if (Platform.OS === 'android') {
        ToastAndroid.show(`${message}`, ToastAndroid.LONG);
      }
      Alert.alert(title, message);
    }
  }
})
```
2. Try sending messages from AppCenter/Push
3. We can't see the alert when the app is opened from the push notification.

### **Details**

1. Which version of the App Center SDK are you using?
    - 1.11.0
2. Which OS version did you experience the issue on?
    - iOS 12.1.2
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - iPhone 6 and iPhone X physical devices
4. What third party libraries are you using?
    - nothing for push notifications
5. Run the following command and paste the output below: `react-native info`

  React Native Environment Info:
    System:
      OS: macOS 10.14.2
      CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
      Memory: 1.45 GB / 16.00 GB
      Shell: 3.2.57 - /bin/bash
    Binaries:
      Node: 8.9.4 - /usr/local/bin/node
      Yarn: 1.1.0 - /usr/local/bin/yarn
      npm: 5.6.0 - /usr/local/bin/npm
      Watchman: 4.9.0 - /usr/local/bin/watchman
    SDKs:
      iOS SDK:
        Platforms: iOS 12.1, macOS 10.14, tvOS 12.1, watchOS 5.1
      Android SDK:
        API Levels: 17, 18, 19, 21, 22, 23, 24, 25, 26, 27, 28
[...]

PR #516, opened 2019-01-28
Update demo app

annakocheshkova:
* [X] Are tests passing locally?
* [X] Did you test your change with either the sample apps that are included in the repository or with a blank app that uses your change?

## Description

Add example how to check if push notification was received in background/foreground to demo app.

## Related PRs or issues

Thanks @jdnichollsc (this comment: https://github.com/Microsoft/AppCenter-SDK-React-Native/issues/505#issuecomment-455622956)

### DIFF
--- a/DemoApp/app/App.js
+++ b/DemoApp/app/App.js
@@ -29,6 +29,7 @@ export default createBottomTabNavigator(
 let previousAppState = AppState.currentState;
 export const setPreviousAppState = (appState) => {
   previousAppState = appState;
+  // Workaround to distinguish between notifications received in the foreground or while the app was in the background.
   setTimeout(() => {
     previousAppState = AppState.currentState;
   }, 1000);
--- a/DemoApp/app/App.js
+++ b/DemoApp/app/App.js
@@ -26,8 +26,21 @@ export default createBottomTabNavigator(
   }
 );
 
+let previousAppState = AppState.currentState;
+export const setPreviousAppState = (appState) => {
+  previousAppState = appState;
+  console.log(appState);
+  // Workaround to distinguish between notifications received in the foreground or while the app was in the background
+  setTimeout(() => {
+    previousAppState = AppState.currentState;
+  }, 1000);
+};
+
 Push.setListener({
   onPushNotificationReceived(pushNotification) {
+    if (previousAppState !== 'active') {
+      Alert.alert('Alert', 'Notification was received on background');
+    }
     let message = pushNotification.message;
     const title = pushNotification.title;
 
--- a/DemoApp/app/App.js
+++ b/DemoApp/app/App.js
@@ -37,9 +37,7 @@ export const setPreviousAppState = (appState) => {
 
 Push.setListener({
   onPushNotificationReceived(pushNotification) {
-    if (previousAppState !== 'active') {
-      Alert.alert('Alert', 'Notification was received in background');
-    }
+    const appStateText = previousAppState === 'active' ? 'foreground' : 'background';
     let message = pushNotification.message;
     const title = pushNotification.title;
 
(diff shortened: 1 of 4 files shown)


DemoApp/app/App.js
```diff
@@ -26,8 +26,20 @@ export default createBottomTabNavigator(
...
+  previousAppState = appState;
+  // Workaround to distinguish between notifications received in the foreground or while the app was in the background.
+  setTimeout(() => {
+    previousAppState = AppState.currentState;
+  }, 1000);
+};
+
 Push.setListener({
   onPushNotificationReceived(pushNotification) {
+    if (previousAppState !== 'active') {
```
dhei: Question - I saw `Alert.alert` call in line 41 and line 71, will you get 2 alerts when receiving push in background?

TestApp/app/App.js
```diff
@@ -26,8 +26,20 @@ export default createBottomTabNavigator(
   }
 );
 
+let previousAppState = AppState.currentState;
+export const setPreviousAppState = (appState) => {
+  previousAppState = appState;
+  // Workaround to distinguish between notifications received in the foreground or while the app was in the background.
```
clpolet: I don't know much in RN, can you please tell me a bit more about this workaround? Also is the choice for 1000ms arbitrary?

DemoApp/app/App.js
dhei: thanks for confirming, combining into one alert sounds good.

TestApp/app/App.js
dhei: @jdnichollsc Can you explain what the purpose of 1000 ms timer? If I understand correctly, you're changing `previousAppState` from active to inactive when push received from background after 1000 ms, why this behavior?

jdnichollsc: @dhei You need a period of time to know if onPushNotificationReceived is running through a notification that arrived when the application was in the background, later you need to update the status for notifications that arrive when the app is active (foreground). 1000 milliseconds is an arbitrary value, the callback should be executed in that time lapse.

clpolet:
Marked as don't merge, we need to experiment a bit more.

clpolet:
We've spent some times discussing this. Relying on the appState to filter notification events isn't reliable except for the very specific scenario of detecting remote notifications received in background via silent push. In the foreground the AC Push callback can be called because either a notification is received or the system notification was tapped by the user but in this case using the appState isn't reliable. That's because in the foreground the appState can either be "inactive" or "active". In the past I've seen people using the "inactive" state to detect the "tap" and the "active" state to detect the "notification received". This can be true most of the time because at the time of the "tap" on the notification the callback will be triggered when the app is in the "inactive" state but the app can be "inactive" for multiple reasons like overlaid by the notification center. If a notification is received but the app is overlaid by the notification center then the appState is inactive and the event is wrongly interpreted as a "tap" instead of a "notification received" event.
Then, Apple introduced the `UNUserNotificationCenterDelegate` in iOS 10 a reliable way to detect the "tap" and "notification received in foreground". I would revisit the code snippet for App Center when React Native will provide easy bridging of `UNUserNotificationCenterDelegate` callbacks.
@jdnichollsc sorry mate, I'll close the PR. This is certainly fitting your use case well but the covered ground is not wide enough to make it official due to the limitations mentioned above. Though I admire your involvement :). Thanks again.

TestApp/app/App.js
clpolet: I personally tend to avoid the use of timer in such situation because the time to achieve the action we are waiting on is relative to many things like the hardware, OS version, state of the devices resources at the moment...I'd rather rely on an event fired at the right time. I know it's not always possible so in the few cases where I can't rely on an event their is also the possibility to do polling (inexpensive frequent check on the state/resource we are waiting for). But it's just my opinion and completely depends on the situation anyway.

jdnichollsc:
No problem, you're welcome guys!

PR #539, opened 2019-02-21
Update Demo App for testing release 1.12.2

UdayaNalla:
## Description

Update Demo App for testing release 1.12.2


guperrot:
This is using the previous version that was not released

DemoApp/package.json
```diff
@@ -8,10 +8,10 @@
     "lint": "./node_modules/.bin/eslint ."
   },
   "dependencies": {
-    "appcenter": "^1.12.0",
-    "appcenter-analytics": "^1.12.0",
-    "appcenter-crashes": "^1.12.0",
-    "appcenter-push": "^1.12.0",
+    "appcenter": "^1.12.1",
```
guperrot: The description says 1.12.2 but this is not the case here.

guperrot:
Also the branch name does not start with `release/` and that's why the CI fails as it tries to eslint a package that is not yet released. Please rename the branch and remake the PR.

PR #548, opened 2019-03-14
Remove linking prompt

dtrabo:

## Description

Modified postlink.js files, deleted any prompts from them

### DIFF
--- a/appcenter-analytics/scripts/postlink.js
+++ b/appcenter-analytics/scripts/postlink.js
@@ -1,28 +1,12 @@
 const rnpmlink = require('appcenter-link-scripts');
-const inquirer = require('inquirer');
 
 // Configure Android first.
 let promise = null;
 if (rnpmlink.android.checkIfAndroidDirectoryExists()) {
     console.log('Configuring AppCenter Analytics for Android');
     promise = rnpmlink.android.initAppCenterConfig()
-        .then(() =>
-            inquirer.prompt([{
-                type: 'list',
-                name: 'whenToEnableAnalytics',
-                message: 'For the Android app, should user tracking be enabled automatically?',
-                choices: [
-                    {
-                        name: 'Enable Automatically',
-                        value: 'ALWAYS_SEND'
-                    },
-                    {
-                        name: 'Enable in JavaScript',
-                        value: 'ENABLE_IN_JS'
-                    }]
-            }])
-        ).then((androidAnswer) => {
-            rnpmlink.android.patchStrings('appCenterAnalytics_whenToEnableAnalytics', androidAnswer.whenToEnableAnalytics);
+        .then(() => {
+            rnpmlink.android.patchStrings('appCenterAnalytics_whenToEnableAnalytics', 'ALWAYS_SEND');
             rnpmlink.android.removeAndroidDuplicateLinks();
         }).catch((e) => {
             console.error(`Could not configure AppCenter Analytics for Android. Error Reason - ${e.message}`);
@@ -39,26 +23,8 @@ if (rnpmlink.ios.checkIfAppDelegateExists()) {
             console.log('Configuring AppCenter Analytics for iOS');
             return rnpmlink.ios.initAppCenterConfig();
         })
-        .then(() =>
-            inquirer.prompt([{
-                type: 'list',
-                name: 'whenToEnableAnalytics',
-                message: 'For the iOS app, should user tracking be enabled automatically?',
-                choices: [
-                    {
-                        name: 'Enable Automatically',
-                        value: 'ALWAYS_SEND'
-                    },
-                    {
-                        name: 'Enable in JavaScript',
-                        value: 'ENABLE_IN_JS'
-                    }]
-            }])
-        )
-        .then((iosAnswer) => {
-            const code = iosAnswer.whenToEnableAnalytics === 'ALWAYS_SEND' ?
-                '  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:true];  // Initialize AppCenter analytics' :
-                '  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:false];  // Initialize AppCenter analytics';
+        .then(() => {
+            const code = '  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:true];  // Initialize AppCenter analytics';
             return rnpmlink.ios.initInAppDelegate('#import <AppCenterReactNativeAnalytics/AppCenterReactNativeAnalytics.h>', code, /.*\[AppCenterReactNativeAnalytics register.*/g);
         })
         .then((file) => {
--- a/appcenter-analytics/scripts/postlink.js
+++ b/appcenter-analytics/scripts/postlink.js
@@ -25,7 +25,6 @@ if (rnpmlink.ios.checkIfAppDelegateExists()) {
         })
         .then(() => {
             const code = '  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:true];  // Initialize AppCenter analytics';
-
             return rnpmlink.ios.initInAppDelegate('#import <AppCenterReactNativeAnalytics/AppCenterReactNativeAnalytics.h>', code, /.*\[AppCenterReactNativeAnalytics register.*/g);
         })
         .then((file) => {
@@ -43,4 +42,3 @@ if (rnpmlink.ios.checkIfAppDelegateExists()) {
             return Promise.resolve();
         });
 }
-return promise;
--- a/appcenter-analytics/scripts/postlink.js
+++ b/appcenter-analytics/scripts/postlink.js
@@ -1,7 +1,7 @@
 const rnpmlink = require('appcenter-link-scripts');
 
 // Configure Android first.
-let promise = null;
+let promise;
 if (rnpmlink.android.checkIfAndroidDirectoryExists()) {
     console.log('Configuring AppCenter Analytics for Android');
     promise = rnpmlink.android.initAppCenterConfig()
--- a/appcenter-crashes/scripts/postlink.js
+++ b/appcenter-crashes/scripts/postlink.js
@@ -1,7 +1,7 @@
 const rnpmlink = require('appcenter-link-scripts');
 
 // Configure Android first.
-let promise = null;
+let promise;
 if (rnpmlink.android.checkIfAndroidDirectoryExists()) {
     console.log('Configuring AppCenter Crashes for Android');
     promise = rnpmlink.android.initAppCenterConfig()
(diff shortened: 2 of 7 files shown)


appcenter-analytics/scripts/postlink.js
```diff
@@ -1,28 +1,12 @@
 const rnpmlink = require('appcenter-link-scripts');
-const inquirer = require('inquirer');
 
 // Configure Android first.
 let promise = null;
```
roland-chernov-akvelon: Nitpick: I think we can remove `return promise` on the last line (45) because it's unused. `promise` should be exported using `module.exports = promise` if we want to get access to it outside of the module.

appcenter-link-scripts/src/android/index.js
```diff
@@ -28,16 +27,10 @@ module.exports = {
             console.log(`Android App Secret is already set in ${config.AppCenterConfigPath}`);
             return Promise.resolve(null);
         }
-        return inquirer.prompt([{
-            type: 'input',
-            message: 'What secret does your Android app use? [None]',
-            name: 'app_secret',
-        }]).then((answers) => {
-            config.set('app_secret', answers.app_secret);
+            config.set('app_secret', 'YOUR_APP_SECRET');
```
roland-chernov-akvelon: Please check indentation. Looks like there is an extra spaces.

appcenter-crashes/scripts/postlink.js
```diff
@@ -1,29 +1,13 @@
 const rnpmlink = require('appcenter-link-scripts');
-const inquirer = require('inquirer');
 
 // Configure Android first.
 let promise = null;
```
roland-chernov-akvelon: The same. Please see my comment above.

appcenter-link-scripts/src/ios/index.js
```diff
@@ -39,23 +38,17 @@ module.exports = {
...
-            }
-        });
+        try {
+            config.set('AppSecret', 'YOUR_APP_SECRET');
+            return config.save()
+                .then((file) => {
+                    console.log(`App Secret for iOS written to ${file}`);
+                    return file;
+                });
+        } catch (e) {
```
roland-chernov-akvelon: Not exactly related to this PR but I think this try/catch doesn't work. Probably we need to use:
```js
            return config.save()
                .then((file) => {
                    console.log(`App Secret for iOS written to ${file}`);
                    return file;
                }).catch((e) => {
                    // catch block
                });
```
Could you please check that?

appcenter-crashes/scripts/postlink.js
```diff
@@ -1,29 +1,13 @@
...
-                    {
-                        name: 'Processed in JavaScript by user',
-                        value: 'ASK_JAVASCRIPT'
-                    }]
-            }])
-        ).then((androidAnswer) => {
+        .then(() => {
             rnpmlink.android.patchStrings('appCenterCrashes_whenToSendCrashes',
-                androidAnswer.whenToSendCrashes);
+                'ALWAYS_SEND');
```
guperrot: This will overwrite the setting that will have to be manually changed by customer.
PatchStrings must be updated to only add the default value the first time.

appcenter-analytics/scripts/postlink.js
```diff
@@ -39,26 +23,8 @@ if (rnpmlink.ios.checkIfAppDelegateExists()) {
...
-                        value: 'ENABLE_IN_JS'
-                    }]
-            }])
-        )
-        .then((iosAnswer) => {
-            const code = iosAnswer.whenToEnableAnalytics === 'ALWAYS_SEND' ?
-                '  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:true];  // Initialize AppCenter analytics' :
-                '  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:false];  // Initialize AppCenter analytics';
+        .then(() => {
+            const code = '  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:true];  // Initialize AppCenter analytics';
```
guperrot: This will overwrite the setting that will have to be manually changed by customer.
This must be added only if no line is there.

appcenter-analytics/scripts/postlink.js
```diff
@@ -1,28 +1,12 @@
...
-                    },
-                    {
-                        name: 'Enable in JavaScript',
-                        value: 'ENABLE_IN_JS'
-                    }]
-            }])
-        ).then((androidAnswer) => {
-            rnpmlink.android.patchStrings('appCenterAnalytics_whenToEnableAnalytics', androidAnswer.whenToEnableAnalytics);
+        .then(() => {
+            rnpmlink.android.patchStrings('appCenterAnalytics_whenToEnableAnalytics', 'ALWAYS_SEND');
```
guperrot: same comment as for crash

appcenter-crashes/scripts/postlink.js
```diff
@@ -40,26 +24,8 @@ if (rnpmlink.ios.checkIfAppDelegateExists()) {
...
-                        value: 'ASK_JAVASCRIPT'
-                    }]
-            }])
-        )
-        .then((iosAnswer) => {
-            const code = iosAnswer.whenToSendCrashes === 'ALWAYS_SEND' ?
-                '  [AppCenterReactNativeCrashes registerWithAutomaticProcessing];  // Initialize AppCenter crashes' :
-                '  [AppCenterReactNativeCrashes register];  // Initialize AppCenter crashes';
+        .then(() => {
+            const code = '  [AppCenterReactNativeCrashes registerWithAutomaticProcessing];  // Initialize AppCenter crashes';
```
guperrot: same comment as for analytics

appcenter-link-scripts/src/ios/index.js
roland-chernov-akvelon: Sorry, nevermind. I've just realized that `AppCenterConfigPlist.save` before returning a `Promise` does synchronous `fs.writeFileSync` that won't be caught by `Promise.catch`.

From my perspective that's a bit unobvious design and for methods which return promises it's better to only use `Promise.reject` so we don't need any try/catch block in addition to `Promise.catch`. But refactoring is probably not in the scope of this task so this PR looks good to me.

appcenter-link-scripts/src/ios/AppDelegate.js
```diff
@@ -35,10 +35,7 @@ AppDelegate.prototype.addInitCode = function (code, oldCodeRegExp) {
     }
     if (this.appDelegateContents.indexOf(code) === -1) {
         /* If new code not found but old code found, replace. */
```
guperrot: Update comment then

appcenter-link-scripts/src/android/index.js
```diff
@@ -28,21 +27,18 @@ module.exports = {
...
+        config.set('app_secret', 'YOUR_APP_SECRET');
+        const file = config.save();
+        console.log(`App Secret for Android written to ${file}`);
+        return Promise.resolve(file);
     },
 
     patchStrings(key, value) {
         const stringsFile = path.join('android', 'app', 'src', 'main', 'res', 'values', 'strings.xml');
         let stringsXml = fs.readFileSync(stringsFile, 'utf-8');
+        if (stringsXml.indexOf(key) > 0) {
```
guperrot: A comment would be nice

appcenter-link-scripts/src/ios/index.js
dhei: @roland-chernov-akvelon you have some great points:
- It's a best practice to use Promise chaining `.then().catch()` instead of mixing Promise with `try {...} catch {...}`.
- The ios `AppCenterConfigPlist.save`(https://github.com/Microsoft/AppCenter-SDK-React-Native/blob/develop/appcenter-link-scripts/src/ios/AppCenterConfig.js#L37) function has a problematic design of asynchronously returning Promise and synchronously throwing errors via `fs.writeFileSync();`.

As the problem is `fs.writeFileSync()` synchronously write to file system, I think we can fix this by calling the asynchronously [`fs.writeFile()`](https://nodejs.org/api/fs.html#fs_fs_writefile_file_data_options_callback).

If it's a small fix, we should just do it as part of this feature.

appcenter-analytics/scripts/postlink.js
dhei: No need to initialize `promise` to `null`. Declaration is enough.

appcenter-crashes/scripts/postlink.js
dhei: No need to initialize `promise` to `null`. Declaration is enough.

appcenter-link-scripts/src/ios/index.js
dtrabo: @dhei Ok, we will do it as separate PR.

dtrabo:
@dhei All scenarios were tested, works fine

dhei:
@dtrabo @MatkovIvan,
[Azure DevOps CI](https://msmobilecenter.visualstudio.com/SDK/_build/results?buildId=539642) is failing as expected as we are in the middle of RN SDK release process. 
There're 2 options, either use a feature branch for all RN linker prompt changes, or wait until RN SDK release is finished. I would suggest let's use a feature branch.

appcenter-analytics/scripts/postlink.js
```diff
@@ -39,26 +23,8 @@ if (rnpmlink.ios.checkIfAppDelegateExists()) {
...
-                        value: 'ENABLE_IN_JS'
-                    }]
-            }])
-        )
-        .then((iosAnswer) => {
-            const code = iosAnswer.whenToEnableAnalytics === 'ALWAYS_SEND' ?
-                '[AppCenterReactNativeAnalytics registerWithInitiallyEnabled:true];  // Initialize AppCenter analytics' :
-                '[AppCenterReactNativeAnalytics registerWithInitiallyEnabled:false];  // Initialize AppCenter analytics';
+        .then(() => {
+            const code = '  [AppCenterReactNativeAnalytics registerWithInitiallyEnabled:true];  // Initialize AppCenter analytics';
```
guperrot: On develop the code was updated to move the 2 spaces management to the core. Please remove them and test on a blank app.

appcenter-crashes/scripts/postlink.js
```diff
@@ -40,26 +24,8 @@ if (rnpmlink.ios.checkIfAppDelegateExists()) {
...
-                        value: 'ASK_JAVASCRIPT'
-                    }]
-            }])
-        )
-        .then((iosAnswer) => {
-            const code = iosAnswer.whenToSendCrashes === 'ALWAYS_SEND' ?
-                '[AppCenterReactNativeCrashes registerWithAutomaticProcessing];  // Initialize AppCenter crashes' :
-                '[AppCenterReactNativeCrashes register];  // Initialize AppCenter crashes';
+        .then(() => {
+            const code = '  [AppCenterReactNativeCrashes registerWithAutomaticProcessing];  // Initialize AppCenter crashes';
```
guperrot: same here remove the 2 spaces

annakocheshkova:
https://github.com/Microsoft/AppCenter-SDK-React-Native/pull/552

Issue #510, opened 2019-01-21
Appcenter installation breaks ios build - 'React/RCTDefines.h' file not found


sanjaypojo:
### **Description**

After installing appcenter by following all the steps in the documentation, android builds perfectly, but ios gives the following error:

```
In file included from ***(path to project root)***/node_modules/appcenter/ios/AppCenterReactNative/AppCenterReactNative.m:

'React/RCTDefines.h' file not found
```

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. `yarn add appcenter appcenter-analytics appcenter-crashes`
2. `react-native link`
3. Fill in android and ios keys
4. Installation completes successfully
5. Clean build folder in Xcode (Cmd + Shift + K)
6. Build in Xcode (Cmd + B)

### **Details**

1. Which version of the App Center SDK are you using?
    - 1.11.0
2. Which OS version did you experience the issue on?
    - iOS 12
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    - iPhone XR emulator
4. What third party libraries are you using?
```
target '****' do
  pod 'AppCenter/Crashes', '~> 1.12.0'
  pod 'AppCenter/Analytics', '~> 1.12.0'
  pod 'AppCenterReactNativeShared', '~> 1.11.0'
  platform :ios, '9.0'
  use_frameworks!
  pod 'React', :path => '../node_modules/react-native', :subspecs => [
      'Core',
      'CxxBridge',
      'RCTImage',
      'RCTNetwork',
      'RCTText',
      'RCTWebSocket',
      'DevSupport',
      'RCTAnimation',
      'RCTLinkingIOS'
  ]
  pod 'react-native-voximplant', path: '../node_modules/react-native-voximplant'
  pod 'yoga', path: '../node_modules/react-native/ReactCommon/yoga'

  # Third party deps podspec link
  pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'

  pod 'ReactNativePermissions', :path => '../node_modules/react-native-permissions'

  pod 'react-native-splash-screen', :path => '../node_modules/react-native-splash-screen'

  pod 'react-native-version-info', :path => '../node_modules/react-native-version-info'

  pod 'react-native-keep-awake', :path => '../node_modules/react-native-keep-awake'

  pod 'RNDeviceInfo', :path => '../node_modules/react-native-device-info'

  # Pods for AppCenter


end
```

[...]

PR #561, opened 2019-03-21
Podspec fix

adkenyon:

## Description

We ran into the build issues after installation, specifically `'React/RCTDefines.h' file not found` mentioned in issue [Issue 510](https://github.com/Microsoft/AppCenter-SDK-React-Native/issues/510). We were able to resolve the issue by modifying the podspec files. Specifically; renamed file, updated spec name, added a dependency for appcenter.podspec (renamed to AppCenterReactNative.podspec), and updated the postlink installation script.

## Related PRs or issues

- [Issue 510](https://github.com/Microsoft/AppCenter-SDK-React-Native/issues/510) Appcenter installation breaks ios build - 'React/RCTDefines.h' file not found

## Misc

 - Changes to the TestApp were huge ~14k files. Those should be included in a separate PR


dhei:
Hey @adkenyon,

Thanks for your PR.

Can you describe step by step how you integrated App Center RN SDK into your app based on this PR? I am curious about your integration steps, especially about whether you use `react-native link` and any pod commands (e.g., `pod install`).

adkenyon:
@dhei 

Here is an example [PodFile](https://github.com/adkenyon/AppCenter-SDK-React-Native/blob/e24db59058738a2abc84e2a240582e92ba1a2b88/TestApp/ios/Podfile) with the changes.

The steps were as follows:
- `react-native link appcenter`
- `react-native link appcenter-analytics`
- `react-native link appcenter-crashes`
- `react-native link appcenter-push`

I don't recall if I ran a pod install in the iOS directory after. I seem to recall react-native link would do it automatically.

dhei:
Hello @adkenyon,

Thanks for contributing and sharing this information. Due to compatibility reasons and not wanting to break existing users, **we cannot accept this PR at this time**.

Integrating a RN iOS library is can take non-trivial effort depends on various conditions. Here's my understanding of the common scenarios of integration any RN iOS library:

1. Automatic integration via `react-native link`. The `react-native link` command injects Pod dependency into Podfile and executes `pod install` under the hood, which manages the library as a pod dependency under `Pod` folder.
- Pros: default and recommended by RN community; easy to on-board.
- Cons: not flexiable of how to manage pod dependency; doesn't work for "brownfield" app.

2. Integrate via `react-native link` by providing npm package local paths in Podfile.
- Pros: more flexiable and you have full control of how dependencies are managed in Podfile; works well for "brownfield" app.
- Cons: manage Podfile ourselves takes more work.

3. Integrate directly via `pod install` by providing npm package local paths in Podfile
- Pros: manages pod dependencies ourselves without `react-native link` complicating things
- Cons: manage Podfile ourselves takes more work; doesn't work for RN.Android.

4. Manual integration without `react-native link` or CocoaPods. You do the manual drag and drop in Xcode.
- Pros & Cons: very error-prone, you should try this only when you have no other options.

To sum up, at this moment App Center RN SDK supports case 1 ([Integrate the SDK automatically](https://docs.microsoft.com/en-us/appcenter/sdk/getting-started/react-native#31-integrate-the-sdk-automatically)), case 3 ([Integrate the iOS SDK without react-native link](https://docs.microsoft.com/en-us/appcenter/sdk/getting-started/react-native#32-ios-only-integrate-the-ios-sdk-without-react-native-link)) and case 4 ([Integrate the iOS SDK manually without react-native link or CocoaPods](https://docs.microsoft.com/en-us/appcenter/sdk/getting-started/react-native#33-ios-only-integrate-the-ios-sdk-manually-without-react-native-link-or-cocoapods)). But not case 2.

If I understand correctly, this PR addresses case 2. However, it seems like this RP would break case 1 that is used by majority of our users, so this doesn't looks like the proper fix. We may re-visit and consider supporting case 2 in the future if more folks want this. In the meanwhile, following case 3 or case 4 should still work.

PR #567, opened 2019-04-03
Add null check for ErrorReport throwable

ascorbic:

## Description
I had an app that was occasionally getting into a state where a pending crash report was causing the app to crash again on launch with a NullPointerException on `getThrowable().getMessage()`. This PR adds a null check for the error report throwable.


dhei:
@ascorbic Thanks for your PR. I am merging this PR and it should be shipped in the next RN SDK release (May 2019).

I have a follow up question, do you have any repro of this bug or at least when does this bug happen? Because App Center RN SDK internally uses App Center Android SDK's API, the NullPointerException is probably thrown from underlying App Center Android SDK. Therefore, we want to fix this bug on Android SDK if possible, and any more information of this bug is appreciated.

ascorbic:
Hi @dhei 
That's great. I have several devices where this currently happens on every launch of the app. Presumably there's a pending crash report on the device, but I can't see anything in the files directory. If you've any tips on how I can get the relevant information from the device then I can see what I can do.

PR #590, opened 2019-06-07
Bootstrap MBAAS Auth Android

jackycodes:

## Description

[AB#62456](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/62456)

Add an empty module for MBAAS Auth Android.


appcenter-auth/android/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+            abiFilters 'armeabi-v7a', 'x86'
+        }
+    }
+}
+
+dependencies {
+    implementation fileTree(include: ['*.jar'], dir: 'libs')
+    //noinspection GradleDynamicVersion
+    implementation 'com.facebook.react:react-native:+'
+    api 'com.microsoft.appcenter:appcenter-auth:2.1.0-7+df4dfd7f7'
```
guperrot:
```suggestion
    api 'com.microsoft.appcenter:appcenter-auth:2.1.0-22+12baf443e'
```
So that you have the new APIs we added (accessToken and idToken in addition to accountId), and UserInformation in the right package.

appcenter-auth/android/src/main/java/com/microsoft/appcenter/reactnative/auth/AppCenterReactNativeAuthModule.java
```diff
@@ -0,0 +1,30 @@
...
+import com.microsoft.appcenter.AppCenter;
+import com.microsoft.appcenter.auth.Auth;
+import com.microsoft.appcenter.reactnative.shared.AppCenterReactNativeShared;
+
+public class AppCenterReactNativeAuthModule extends BaseJavaModule {
+
+    public AppCenterReactNativeAuthModule(Application application) {
+        AppCenterReactNativeShared.configureAppCenter(application);
+        if (AppCenter.isConfigured()) {
+            AppCenter.start(Auth.class);
```
guperrot: From https://docs.microsoft.com/en-us/appcenter/sdk/auth/android

> The Auth module must be started at the same time as other modules, in a AppCenter.start call. If you decide to start the Auth service separately using AppCenter.startService, the app will not sign in automatically after the restart and the user will have to call signIn again.

We might need to start Auth from react-native shared (in configureAppCenter) using reflection or think about another solution.

Does not have to be fixed for this particular bootstrap PR but please add a work item to track that issue.
It's the same for iOS.

appcenter-auth/android/build.gradle
jackycodes: Changed

jackycodes:
> We should only have 1 base branch mbaas instead of 1 for auth and 1 for data.
> Only 1 comment otherwise.

Having separate branches allows us to merge Auth to develop faster since Auth does not depend on Data.

TestApp/android/app/src/main/java/com/testapp/MainApplication.java
```diff
@@ -18,6 +18,7 @@
 import com.microsoft.appcenter.reactnative.appcenter.AppCenterReactNativePackage;
 import com.microsoft.appcenter.reactnative.crashes.AppCenterReactNativeCrashesPackage;
 import com.microsoft.appcenter.reactnative.push.AppCenterReactNativePushPackage;
+import com.microsoft.appcenter.reactnative.auth.AppCenterReactNativeAuthPackage;
```
dhei: sort imports.

appcenter-auth/.eslintrc
```diff
@@ -0,0 +1,30 @@
+{
```
dhei: Please add `appcenter-auth` directory to https://github.com/microsoft/appcenter-sdk-react-native/blob/develop/scripts/eslint.sh#L7 so that CI will run eslint on `appcenter-auth` directory.

We should do the same for `appcenter-data` module, but that can be done by a separate PR.

jackycodes: I'll do a separate PRs to update the script files

appcenter-auth/android/src/main/java/com/microsoft/appcenter/reactnative/auth/AppCenterReactNativeAuthModule.java
jackycodes: Sounds good

TestApp/android/app/src/main/java/com/testapp/MainApplication.java
jackycodes: Done

PR #600, opened 2019-06-17
[TestApp] Fix warning of `Async Storage` from RN 0.59

russelarms:

## Description

React Native 0.59.x caused a warning in TestApp UI that `Async Storage` has been extracted from react-native core and moved to an independent module.

Let's fix this warning in TestApp and DemoApp.

## Related PRs or issues

[AB#63182](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/63182)


DemoApp/android/app/src/main/AndroidManifest.xml
```diff
@@ -6,9 +6,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE"/>
-    <uses-sdk
-        android:minSdkVersion="16"
```
jwallra: I don't think we can change the minimum version for our test app until we've stopped supporting the lower versions

jwallra: Is it a requirement of the new async storage package?

TestApp/jest/setup.js
```diff
@@ -2,4 +2,8 @@
 // Licensed under the MIT License.
 
 // https://github.com/expo/expo/issues/1705#issuecomment-385338032
```
jwallra: This comment needs to be moved now since you inserted an import

DemoApp/android/app/src/main/AndroidManifest.xml
guperrot: It should come from gradle the min and target versions, please do not use Manifest to manage that.

russelarms: It was requested by the build system:
```
ERROR: The minSdk version should not be declared in the android manifest file. You can move the version from the manifest to the defaultConfig in the build.gradle file.
Remove minSdkVersion and sync project
```
I'll remove it from Manifest.

TestAppTypescript/src/CrashesScreen.tsx
```diff
@@ -60,7 +60,7 @@ export default class CrashesScreen extends Component<
       component.setState({ lastSessionStatus: status });
     }
 
-    const textAttachmentValue = await AttachmentsProvider.getTextAttachment();
+    const textAttachmentValue = await AttachmentsProvider.getTextAttachment() || "";
```
dhei: Isn't `getTextAttachment()`(https://github.com/microsoft/appcenter-sdk-react-native/blob/develop/TestAppTypescript/src/AttachmentsProvider.ts#L68) already returning empty string when it can't get a text attachment? Why do we need empty string here?

TestApp/package.json
```diff
@@ -8,6 +8,7 @@
     "lint": "./node_modules/.bin/eslint ."
   },
   "dependencies": {
+    "@react-native-community/async-storage": "^1.5.0",
```
jwallra: @dhei Should we consider pinning this version or is it OK to allow newer ones?

TestAppTypescript/android/app/build.gradle
```diff
@@ -146,8 +146,8 @@ dependencies {
     implementation fileTree(dir: "libs", include: ["*.jar"])
     implementation "com.android.support:appcompat-v7:${rootProject.ext.supportLibVersion}"
     implementation "com.facebook.react:react-native:+"  // From node_modules
-    implementation "com.google.firebase:firebase-core:16.0.1"
-    implementation 'com.google.firebase:firebase-messaging:17.0.0'
+    implementation "com.google.firebase:firebase-core:16.0.9"
```
guperrot: you can also remove it altogether now

TestAppTypescript/android/build.gradle
```diff
@@ -14,7 +14,7 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.3.1'
-        classpath 'com.google.gms:google-services:4.0.1'
+        classpath 'com.google.gms:google-services:4.1.0'
```
guperrot: did you validate push token is sent?

TestApp/package.json
dhei: I don't have a strong preference. NPM's default behavior is to use highest minor version (no pinning) and sometimes minor versions contain breaking change therefore breaking the app. 

It's probably okay not pinning version and use `^` for external dependencies except `react` and `react-native`.

TestAppTypescript/android/build.gradle
dhei: +1 for testing Push every time we bump firebase dependencies.

TestAppTypescript/src/CrashesScreen.tsx
russelarms: As `AsyncStorage.getItem` has `Promise<string | null>` return type, typescript compiler returns an error:
`Type 'string | null' is not assignable to type 'string’.`
I removed that line and updated `AttachmentsProvider.getItemFromStorage` method to handle this type check. Please, check it out.

TestAppTypescript/android/build.gradle
russelarms: I've rolled back this change as it were firebase dependencies that were causing build issues with TestAppTypescript.

TestAppTypescript/android/app/build.gradle
```diff
@@ -73,7 +73,7 @@ import com.android.build.OutputFile
  */
 
 project.ext.react = [
-    entryFile: "index.js"
+        entryFile: "index.js"
```
dhei:
```suggestion
    entryFile: "index.js"
```
Fix indentation.

TestApp/package.json
dhei: @jwallra Btw, thanks for bringing this up. We can revisit this decision if we run into problems of not pinning version.

Issue #598, opened 2019-06-14
UI API called on background thread


danfri86:
AppCenter version: 2.0.0
iOS version: Iphone X 11.4 & 12.2 and iPhone 8 11.4 & 12.2. Haven't tested on more.
Error exists on both simulator and physical device.

I'm delaying the permission dialog by setting the 'EnablePushInJavascript' in AppCenter-Config.plist. It's working.

Later during my onboarding-process I call Push.setEnabled(true) which also works and the permission dialog shows. However I get an error:

_"Bragd dev" below is my app name in development_
```
Main Thread Checker: UI API called on a background thread: -[UIApplication registerForRemoteNotifications]
PID: 25527, TID: 1359966, Thread name: (none), Queue name: com.facebook.react.AppCenterReactNativePushQueue, QoS: 0
Backtrace:
4   Bragd dev                           0x00000001061e7dfb -[MSPush registerForRemoteNotifications] + 344
5   Bragd dev                           0x00000001061e7ac0 -[MSPush applyEnabledState:] + 335
6   Bragd dev                           0x00000001061e7678 -[MSPush startWithChannelGroup:appSecret:transmissionTargetToken:fromApplication:] + 110
7   Bragd dev                           0x000000010614ea3c -[MSAppCenter startService:withAppSecret:transmissionTargetToken:andSendLog:fromApplication:] + 1087
8   Bragd dev                           0x000000010614d2cf +[MSAppCenter startService:] + 192
9   Bragd dev                           0x0000000106704fca -[AppCenterReactNativePush setEnabled:resolver:rejecter:] + 154
10  CoreFoundation                      0x000000010d39bccc __invoking___ + 140
11  CoreFoundation                      0x000000010d39bb84 -[NSInvocation invoke] + 308
12  CoreFoundation                      0x000000010d3b48d6 -[NSInvocation invokeWithTarget:] + 54
13  Bragd dev                           0x0000000106344a7a -[RCTModuleMethod invokeWithBridge:module:arguments:] + 2810
14  Bragd dev                           0x00000001063e2136 _ZN8facebook5reactL11invokeInnerEP9RCTBridgeP13RCTModuleDatajRKN5folly7dynamicE + 790
15  Bragd dev                           0x00000001063e1c43 _ZZN8facebook5react15RCTNativeModule6invokeEjON5folly7dynamicEiENK3$_0clEv + 131
[...]

PR #603, opened 2019-06-20
Wording in 1 change

guperrot:

## Description

Improve changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Version 2.1.0 of the App Center React Native SDK includes a new module: Auth.
 
 ### iOS
 
-* **[Fix]** Fix a crash that could occur when delaying the start of the Push module.
+* **[Fix]** Fix registering push notifications in the UI thread when delaying the start of the Push module.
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ Version 2.1.0 of the App Center React Native SDK includes a new module: Auth.
 
 ### iOS
 
-* **[Fix]** Fix calling UI API on background thread.
+* **[Fix]** Fix a crash that could occur when delaying the start of the Push module.
```
dhei: The bug reported in https://github.com/microsoft/appcenter-sdk-react-native/issues/598, I don't see "crash" is mentioned anywhere in the Issue. 

I am curious if this bug would actually cause the app to crash. 

Maybe I am being picky here, if this bug doesn't crash the app, we can rephrase the wording to something like this:
```
Fix a bug where SDK calls UI API on a background thread when delaying the start of the Push module.
```

guperrot: Ok, but now I think the bug should have been fixed on native side.

PR #608, opened 2019-06-22
Merge to Develop for Release 2.1.0

v-eacaoi:
## Description

Updating develop to version 2.1.0


TestApp/android/app/src/main/java/com/testapp/MainApplication.java
```diff
@@ -7,6 +7,11 @@
 import android.util.Log;
```
guperrot: This file needs to be reverted

TestApp/ios/Podfile.lock
```diff
@@ -8,7 +8,7 @@ PODS:
     - AppCenter/Core
   - AppCenter/Push (2.0.2-82):
     - AppCenter/Core
-  - AppCenterReactNativeShared (2.0.0):
+  - AppCenterReactNativeShared (2.1.0):
```
guperrot: We need to clean the -82 thing, everything should be 2.1.0 here

newapp/android/app/BUCK
```diff
@@ -0,0 +1,55 @@
+# To learn about Buck see [Docs](https://buckbuild.com/).
```
guperrot: This newapp should not have been committed

PR #614, opened 2019-06-25
Implement `create` on Android

jackycodes:

## Description
[AB#62463](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/62463)
- Implemented `create` for React Native Android.
- Added `WriteOptions` for `create`
- Created helper functions, `convertReadableMapToJsonObject` and `convertReadableArrayToJsonArray` to convert JS to Java JSON Objects
- Create a private `Consumer` class that implements `AppCenterConsumer` to handle callbacks for `read` and create`

### DIFF
--- a/appcenter-data/Data.js
+++ b/appcenter-data/Data.js
@@ -40,7 +40,7 @@ function read(documentId, partition, readOptions) {
 }
 
 /**
- * Create a document
+ * Create a document.
  *
  * @param {string} documentId - The CosmosDB document id.
  * @param {object} document - The document.
@@ -69,7 +69,7 @@ Data.WriteOptions = class {
 };
 
 function convertTimestampToDate(result) {
-    if (!result === undefined || !result.lastUpdatedDate === undefined) {
+    if (result === undefined || result.lastUpdatedDate === undefined) {
         return result;
     }
     // Create a new `Date` object from timestamp as milliseconds.
--- a/appcenter-data/Data.js
+++ b/appcenter-data/Data.js
@@ -69,6 +69,9 @@ Data.WriteOptions = class {
 };
 
 function convertTimestampToDate(result) {
+    if (!result === undefined || !result.lastUpdatedDate === undefined) {
+        return result;
+    }
     // Create a new `Date` object from timestamp as milliseconds.
     result.lastUpdatedDate = new Date(result.lastUpdatedDate);
     return result;
--- a/appcenter-data/Data.js
+++ b/appcenter-data/Data.js
@@ -69,7 +69,7 @@ Data.WriteOptions = class {
 };
 
 function convertTimestampToDate(result) {
-    if (result === undefined || result.lastUpdatedDate === undefined) {
+    if (!result || !result.lastUpdatedDate) {
         return result;
     }
     // Create a new `Date` object from timestamp as milliseconds.
--- a/appcenter-data/Data.js
+++ b/appcenter-data/Data.js
@@ -19,19 +19,42 @@ const DefaultPartitions = {
 const Data = {
     TimeToLive,
     DefaultPartitions,
-    read(documentId, partition, readOptions) {
-        if (readOptions === undefined) {
-            readOptions = new Data.ReadOptions(TimeToLive.DEFAULT);
-        }
-        return AppCenterReactNativeData.read(documentId, partition, readOptions).then(convertTimestampToDate);
-    },
-    create(documentId, document, partition, writeOptions) {
-        if (writeOptions === undefined) {
-            writeOptions = new Data.WriteOptions(TimeToLive.DEFAULT);
-        }
-        return AppCenterReactNativeData.create(documentId, document, partition, writeOptions).then(convertTimestampToDate);
-    }
+    read,
+    create
 };
+
+/**
+ * Read a document.
+ *
+ * @param {string} documentId - The CosmosDB document id.
+ * @param {string} partition - The CosmosDB partition key.
+ * @param {object} readOptions - Cache read options when the operation is done offline.
+ * @return {Promise} Future asynchronous operation with result being the document with metadata.
+ * If the operation fails, the error can be checked by reading the error object.
+ */
+function read(documentId, partition, readOptions) {
+    if (readOptions === undefined) {
+        readOptions = new Data.ReadOptions(TimeToLive.DEFAULT);
+    }
+    return AppCenterReactNativeData.read(documentId, partition, readOptions).then(convertTimestampToDate);
+}
+
+/**
+ * Create a document
+ *
+ * @param {string} documentId - The CosmosDB document id.
+ * @param {object} document - The document.
+ * @param {string} partition - The CosmosDB partition key.
+ * @param {object} writeOptions - Cache write options when the operation is done offline.
+ * @return {Promise} Future asynchronous operation with result being the document with metadata.
+ * If the operation fails, the error can be checked by reading the error object.
+ */
+function create(documentId, document, partition, writeOptions) {
+    if (writeOptions === undefined) {
+        writeOptions = new Data.WriteOptions(TimeToLive.DEFAULT);
+    }
+    return AppCenterReactNativeData.create(documentId, document, partition, writeOptions).then(convertTimestampToDate);
+}
 
 Data.ReadOptions = class {
     constructor(timeToLive) {
@@ -46,9 +69,11 @@ Data.WriteOptions = class {
 };
 
 function convertTimestampToDate(result) {
-    // Create a new `Date` object from timestamp as milliseconds
+    // Create a new `Date` object from timestamp as milliseconds.
     result.lastUpdatedDate = new Date(result.lastUpdatedDate);
     return result;
 }
+
+this.read();
 
 module.exports = Data;
--- a/appcenter-data/Data.js
+++ b/appcenter-data/Data.js
@@ -23,11 +23,10 @@ const Data = {
         if (readOptions === undefined) {
             readOptions = new Data.ReadOptions(TimeToLive.DEFAULT);
         }
-        return AppCenterReactNativeData.read(documentId, partition, readOptions).then((result) => {
-            // Create a new `Date` object from timestamp as milliseconds
-            result.lastUpdatedDate = new Date(result.lastUpdatedDate);
-            return result;
-        });
+        return AppCenterReactNativeData.read(documentId, partition, readOptions).then(convertTimestampToDate);
+    },
+    create(documentId, document, partition) {
+        return AppCenterReactNativeData.create(documentId, document, partition).then(convertTimestampToDate);
     }
 };
 
@@ -36,5 +35,11 @@ Data.ReadOptions = class {
         this.timeToLive = timeToLive;
     }
 };
+
+function convertTimestampToDate(result) {
+    // Create a new `Date` object from timestamp as milliseconds
+    result.lastUpdatedDate = new Date(result.lastUpdatedDate);
+    return result;
+}
 
 module.exports = Data;
--- a/appcenter-data/Data.js
+++ b/appcenter-data/Data.js
@@ -25,12 +25,21 @@ const Data = {
         }
         return AppCenterReactNativeData.read(documentId, partition, readOptions).then(convertTimestampToDate);
     },
-    create(documentId, document, partition) {
-        return AppCenterReactNativeData.create(documentId, document, partition).then(convertTimestampToDate);
+    create(documentId, document, partition, writeOptions) {
+        if (writeOptions === undefined) {
+            writeOptions = new Data.WriteOptions(TimeToLive.DEFAULT);
+        }
+        return AppCenterReactNativeData.create(documentId, document, partition, writeOptions).then(convertTimestampToDate);
     }
 };
 
 Data.ReadOptions = class {
+    constructor(timeToLive) {
+        this.timeToLive = timeToLive;
+    }
+};
+
+Data.WriteOptions = class {
     constructor(timeToLive) {
         this.timeToLive = timeToLive;
     }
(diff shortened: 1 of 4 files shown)


appcenter-data/Data.js
```diff
@@ -37,4 +39,16 @@ Data.ReadOptions = class {
...
 };
 
+Data.WriteOptions = class {
+    constructor(timeToLive) {
+        this.timeToLive = timeToLive;
+    }
+};
+
+function convertTimestampToDate(result) {
+    // Create a new `Date` object from timestamp as milliseconds
```
guperrot:
```suggestion
    // Create a new `Date` object from timestamp as milliseconds.
```

appcenter-data/Data.js
```diff
@@ -23,11 +23,13 @@ const Data = {
...
             readOptions = new Data.ReadOptions(TimeToLive.DEFAULT);
         }
-        return AppCenterReactNativeData.read(documentId, partition, readOptions).then((result) => {
-            // Create a new `Date` object from timestamp as milliseconds
-            result.lastUpdatedDate = new Date(result.lastUpdatedDate);
-            return result;
-        });
+        return AppCenterReactNativeData.read(documentId, partition, readOptions).then(convertTimestampToDate);
+    },
+    create(documentId, document, partition, writeOptions) {
```
guperrot: Can you test how we can write "Javadoc"? Would it be visible in VS code when mousing over?

appcenter-data/Data.js
```diff
@@ -37,4 +39,16 @@ Data.ReadOptions = class {
...
 
+Data.WriteOptions = class {
+    constructor(timeToLive) {
+        this.timeToLive = timeToLive;
+    }
+};
+
+function convertTimestampToDate(result) {
+    // Create a new `Date` object from timestamp as milliseconds
+    result.lastUpdatedDate = new Date(result.lastUpdatedDate);
```
dhei: Question: Can `result` be undefined or null? Do we handle falsy values gracefully here?

appcenter-data/Data.js
jackycodes: Fixed

appcenter-data/Data.js
jackycodes: Added it looks something like this 
Screen Shot 2019-06-26 at 2 58 44 PM

appcenter-data/Data.js
```diff
@@ -46,9 +69,11 @@ Data.WriteOptions = class {
 };
 
 function convertTimestampToDate(result) {
-    // Create a new `Date` object from timestamp as milliseconds
+    // Create a new `Date` object from timestamp as milliseconds.
     result.lastUpdatedDate = new Date(result.lastUpdatedDate);
     return result;
 }
 
+this.read();
```
guperrot: What is this?

appcenter-data/Data.js
```diff
@@ -19,20 +19,43 @@ const DefaultPartitions = {
...
+ */
+function read(documentId, partition, readOptions) {
+    if (readOptions === undefined) {
+        readOptions = new Data.ReadOptions(TimeToLive.DEFAULT);
+    }
+    return AppCenterReactNativeData.read(documentId, partition, readOptions).then(convertTimestampToDate);
+}
+
+/**
+ * Create a document
```
guperrot: Add dots at end of comments.

jackycodes: Added

appcenter-data/Data.js
jackycodes: It shouldn't be since I pass this function as a callback for a success. I added some error handling code just in to be safe.

appcenter-data/Data.js
```diff
@@ -69,6 +69,9 @@ Data.WriteOptions = class {
 };
 
 function convertTimestampToDate(result) {
+    if (!result === undefined || !result.lastUpdatedDate === undefined) {
```
michaelsmitha: If `result` is either `null` or `undefined`, `!result` will return `true`, because `null` and `undefined` are falsey values.
`true` will never be `===` to `undefined`. same issue with `lastUpdateDate`

appcenter-data/Data.js
```diff
@@ -69,7 +69,7 @@ Data.WriteOptions = class {
 };
 
 function convertTimestampToDate(result) {
-    if (!result === undefined || !result.lastUpdatedDate === undefined) {
+    if (result === undefined || result.lastUpdatedDate === undefined) {
```
michaelsmitha: Doesn't handle `null` case, if that needs handling.

appcenter-data/Data.js
jackycodes: Oops, fixed

appcenter-data/Data.js
jackycodes: Would `!result` take care of both?

michaelsmitha: Yep, `(!result || !result.lastUpdatedDate)`

jackycodes: Awesome, added

appcenter-data/Data.js
```diff
@@ -19,22 +19,62 @@ const DefaultPartitions = {
...
+
+/**
+ * Create a document.
+ *
+ * @param {string} documentId - The CosmosDB document id.
+ * @param {object} document - The document.
+ * @param {string} partition - The CosmosDB partition key.
+ * @param {object} writeOptions - Cache write options when the operation is done offline.
+ * @return {Promise} Future asynchronous operation with result being the document with metadata.
+ * If the operation fails, the error can be checked by reading the error object.
```
guperrot: no it throws an exception on Javascript.

appcenter-data/Data.js
```diff
@@ -19,22 +19,62 @@ const DefaultPartitions = {
...
 };
 
+/**
+ * Read a document.
+ *
+ * @param {string} documentId - The CosmosDB document id.
+ * @param {string} partition - The CosmosDB partition key.
+ * @param {object} readOptions - Cache read options when the operation is done offline.
+ * @return {Promise} Future asynchronous operation with result being the document with metadata.
+ * If the operation fails, the error can be checked by reading the error object.
```
guperrot: it throws an exception

appcenter-data/android/src/main/java/com/microsoft/appcenter/reactnative/data/AppCenterReactNativeDataModule.java
```diff
@@ -67,48 +68,68 @@ public void read(String documentId, String partition, ReadableMap readOptionsMap
...
+        }
+        JsonObject jsonObject = AppCenterReactNativeDataUtils.convertReadableMapToJsonObject(readableMap);
+        Data.create(documentId, jsonObject, JsonElement.class, partition, writeOptions).thenAccept(new Consumer("Create failed", promise));
+    }
+
+    private class Consumer implements AppCenterConsumer<DocumentWrapper<JsonElement>> {
+        private Promise promise;
+        private String errorCode;
+
+        private Consumer(String errorCode, final Promise promise) {
```
guperrot:
```suggestion
        private Consumer(String errorCode, Promise promise) {
```

appcenter-data/android/src/main/java/com/microsoft/appcenter/reactnative/data/AppCenterReactNativeDataModule.java
```diff
@@ -67,48 +68,68 @@ public void read(String documentId, String partition, ReadableMap readOptionsMap
...
+            jsDocumentWrapper.putString(PARTITION_KEY, documentWrapper.getPartition());
+            if (documentWrapper.getError() != null) {
+                DataException dataException = documentWrapper.getError();
+                promise.reject(errorCode, dataException.getMessage(), dataException, jsDocumentWrapper);
+                return;
+            }
+            JsonElement deserializedValue = documentWrapper.getDeserializedValue();
+            jsDocumentWrapper.putString(JSON_VALUE_KEY, documentWrapper.getJsonValue());
+
+            // Pass milliseconds back to JS object since `WritableMap` does not support `Date` as values
```
guperrot: We use /* comments on Java

guperrot: And dots

appcenter-data/Data.js
jackycodes: > If the operation fails, an exception is thrown and the promise is rejected. The exception can be viewed by reading the error object.

How does that sound?

guperrot: If the operation fails, the promise is rejected with an exception containing the details of the error.

appcenter-data/android/src/main/java/com/microsoft/appcenter/reactnative/data/AppCenterReactNativeDataModule.java
guperrot: We also put new lines between fields, and on Android we usually use mField convention (and thus not `this`).

appcenter-data/android/src/main/java/com/microsoft/appcenter/reactnative/data/AppCenterReactNativeDataUtils.java
```diff
@@ -64,4 +69,67 @@ public static WritableArray convertJsonArrayToWritableArray(JsonArray jsonArray)
...
+        if (readableMap == null) {
+            return jsonObject;
+        }
+        ReadableMapKeySetIterator iterator = readableMap.keySetIterator();
+        while (iterator.hasNextKey()) {
+            String key = iterator.nextKey();
+            ReadableType type = readableMap.getType(key);
+            switch (type) {
+                case Number:
+                    jsonObject.addProperty(key, readableMap.getDouble(key));
```
dhei: I had an offline discussion with @jackycodes and @michaelsmitha on handling of different types of numbers (integer/long/double) here. 

Note that this line parse a JavaScript `number` value into Java `Double` value. **The conclusion is it is fine to do so because JSON doesn't care about integer/long/double types and treats them as JavaScript `number`.** Cosmos DB saves document as JSON. No matter it is integer `123`, long `123` or double `123.45`, it's treated as a JavaScript `number` in JSON.

An interesting edge case is `number` with trailing zeros (e.g., `123.00`) is considered as strictly equal (`===`) with the same number without trailing zeros (e.g., `123`). We verified Cosmos DB actually strips trailing zeros on saving the JSON document. In other words, the JSON documents in Cosmos would never have number properties with trailing zeros (e.g., `123.00`).

For example, `create` API scenarios like:
- JSON document property `123` would be saved as `123` in cosmos.
- JSON document property `123.45` would be saved as `123.45` in cosmos.
- JSON document property `123.00` would be saved as `123` in cosmos because of JSON not preserving trailing zeros.

appcenter-data/Data.js
jackycodes: This is committed by accident. It's removed.

appcenter-data/Data.js
jackycodes: Changed

appcenter-data/Data.js
jackycodes: Changed

appcenter-data/android/src/main/java/com/microsoft/appcenter/reactnative/data/AppCenterReactNativeDataModule.java
jackycodes: Added

appcenter-data/android/src/main/java/com/microsoft/appcenter/reactnative/data/AppCenterReactNativeDataModule.java
jackycodes: Changed

appcenter-data/android/src/main/java/com/microsoft/appcenter/reactnative/data/AppCenterReactNativeDataUtils.java
```diff
@@ -64,4 +69,68 @@ public static WritableArray convertJsonArrayToWritableArray(JsonArray jsonArray)
...
+                    jsonArray.add(readableArray.getBoolean(i));
+                    break;
+                case Map:
+                    jsonArray.add(convertReadableMapToJsonObject(readableArray.getMap(i)));
+                    break;
+                case Array:
+                    jsonArray.add(convertReadableArrayToJsonArray(readableArray.getArray(i)));
+                    break;
+                default:
+                    jsonArray.add(JsonNull.INSTANCE);
```
jackycodes: Currently, the native Android SDK does not support serialization of null fields. A separate work item will be created to update the native SDK to enable this feature.

jackycodes:
> So though having a testapp UI would be too big for this PR, we should at least have hardcoded Data calls that we can check in logcat like we did for Xamarin when lacking time to build a Data UI.
> 
> For example create a document with random id, and read it after sign in (as a sign in post action). This should be a complex object that test a bit of every type.

Added

appcenter-data/Data.js
```diff
@@ -19,22 +19,62 @@ const DefaultPartitions = {
...
 };
 
+Data.WriteOptions = class {
+    constructor(timeToLive) {
+        this.timeToLive = timeToLive;
+    }
+};
+
+function convertTimestampToDate(result) {
+    if (!result || !result.lastUpdatedDate) {
```
dhei: This code can be simplified with one `return`:
```
if (result && result.lastUpdatedDate) {
    result.lastUpdatedDate = new Date(result.lastUpdatedDate);
}
return result;
```

jackycodes: Sounds good, I'll put this up for the next PR

PR #617, opened 2019-06-27
Fix Authentication status to "Authentication status unknown" when app is launched

charmian01:

* [X] Are tests passing locally?

## Description

Displayed "Authentication status unknown" when app is launched.
Changed "Is Token Set" value to "User is not authenticated", "User is authenticated".

## Related PRs or issues

Bug [63639](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/63639/?triage=true)

### DIFF
--- a/DemoApp/app/screens/AppCenterScreen.js
+++ b/DemoApp/app/screens/AppCenterScreen.js
@@ -77,7 +77,7 @@ export default class AppCenterScreen extends Component {
     startupMode: StartupModes[0],
     userId: '',
     accountId: '',
-    isTokenSet: false
+    isTokenSet: 'Authentication status unknown'
   }
 
   async componentWillMount() {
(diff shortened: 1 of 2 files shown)


DemoApp/app/screens/AppCenterScreen.js
```diff
@@ -77,7 +77,7 @@ export default class AppCenterScreen extends Component {
     startupMode: StartupModes[0],
     userId: '',
     accountId: '',
-    isTokenSet: false
+    isTokenSet: 'Authentication status unknown'
```
MatkovIvan: Since if it isn't bool value anymore - please rename it

charmian01: Renamed it.

PR #632, opened 2019-07-08
Use working jest option

michaelsmitha:

## Description

Get Jest unit tests running on the SDK

image

## Related PRs or issues

[AB#64086](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/64086)


appcenter-analytics/package.json
```diff
@@ -43,9 +43,7 @@
   },
   "jest": {
     "preset": "react-native",
-    "setupFilesAfterEnv": [
-      "./jest/setup.js"
-    ],
+    "setupTestFrameworkScriptFile": "./jest/setup.js",
```
dhei: Looks like [setupTestFrameworkScriptFile is deprecated in favor of setupFilesAfterEnv](https://jestjs.io/docs/en/configuration#setupfilesafterenv-array).

michaelsmitha: @dhei `setupFilesAfterEnv` doesn't work with our current setup. I know @jackycodes was looking at it a while ago, I don't think he ever got it working.

dhei: Do we know when jest will deprecate `setupTestFrameworkScriptFile`?

michaelsmitha: @dhei no, but it looks like jest@^24.x.x fixes the issue. I'll upgrade jest instead - that's better.

michaelsmitha:
@dhei done. I had to bump the `devDependencies` versions of `react, `react-native`,  and `jest` - I just made them the same versions that are in `appcenter-analytics` though

PR #633, opened 2019-07-09
Update android TestApp for autolinking

dhei:

## Description

TestApp Android is now on RN 0.60.0 and uses autolinking.

I did smoke testing on analytics/crashes/push/auth modules and they all look good.

### DIFF
--- a/TestApp/android/app/src/main/java/com/testapp/MainApplication.java
+++ b/TestApp/android/app/src/main/java/com/testapp/MainApplication.java
@@ -6,24 +6,13 @@ package com.testapp;
 import android.app.Application;
 import android.util.Log;
 
+import com.facebook.react.PackageList;
 import com.facebook.react.ReactApplication;
-import com.swmansion.gesturehandler.react.RNGestureHandlerPackage;
 import com.facebook.react.ReactNativeHost;
-import com.reactnativecommunity.asyncstorage.AsyncStoragePackage;
 import com.facebook.react.ReactPackage;
-import com.facebook.react.shell.MainReactPackage;
 import com.facebook.soloader.SoLoader;
-import com.imagepicker.ImagePickerPackage;
 import com.microsoft.appcenter.AppCenter;
-import com.microsoft.appcenter.reactnative.analytics.AppCenterReactNativeAnalyticsPackage;
-import com.microsoft.appcenter.reactnative.appcenter.AppCenterReactNativePackage;
-import com.microsoft.appcenter.reactnative.auth.AppCenterReactNativeAuthPackage;
-import com.microsoft.appcenter.reactnative.crashes.AppCenterReactNativeCrashesPackage;
-import com.microsoft.appcenter.reactnative.push.AppCenterReactNativePushPackage;
 
-import com.rnfs.RNFSPackage;
-
-import java.util.Arrays;
 import java.util.List;
 
 public class MainApplication extends Application implements ReactApplication {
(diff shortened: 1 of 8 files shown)


TestApp/update-npm-packages.sh
```diff
@@ -24,5 +24,5 @@ npm install
 echo "Build shared framework..."
 (cd ../AppCenterReactNativeShared/ios && SRCROOT=`pwd` ./build-fat-framework.sh)
 
-echo "Running jetify for react-native 0.60.0-rc3..."
+echo "Running jetify for react-native 0.60.0..."
```
guperrot: We will update RN and forget to update this echo line. Can we remove version or why we do it?
```suggestion
echo "Running jetify"
```

TestApp/android/app/src/main/java/com/testapp/MainApplication.java
```diff
@@ -37,19 +26,14 @@ public boolean getUseDeveloperSupport() {
...
-                    new AppCenterReactNativeAuthPackage(getApplication()),
-                    new AppCenterReactNativePushPackage(getApplication()),
-                    new AppCenterReactNativeCrashesPackage(getApplication(), getResources().getString(R.string.appCenterCrashes_whenToSendCrashes)),
-                    new AppCenterReactNativeAnalyticsPackage(getApplication(), getResources().getString(R.string.appCenterAnalytics_whenToEnableAnalytics)),
-                    new AppCenterReactNativePackage(getApplication()),
-                    new RNFSPackage(),
-                    new ImagePickerPackage(),
-                    new AsyncStoragePackage()
-            );
+            @SuppressWarnings("UnnecessaryLocalVariable")
```
guperrot:
```suggestion

            @SuppressWarnings("UnnecessaryLocalVariable")
```

TestApp/android/app/src/main/java/com/testapp/MainApplication.java
```diff
@@ -37,19 +26,14 @@ public boolean getUseDeveloperSupport() {
...
+            List<ReactPackage> packages = new PackageList(this).getPackages();
```
guperrot: inline variable

dhei: This is code generated by `react-native init`, so probably okay to keep it.

TestApp/android/app/src/main/java/com/testapp/MainApplication.java
```diff
@@ -37,19 +26,15 @@ public boolean getUseDeveloperSupport() {
 
         @Override
         protected List<ReactPackage> getPackages() {
-            return Arrays.asList(
-                    new MainReactPackage(),
-                    new RNGestureHandlerPackage(),
-                    new TestAppNativePackage(getApplication()),
-                    new AppCenterReactNativeAuthPackage(getApplication()),
```
dhei: Currently we have a limitation that `AppCenterReactNativeAuthPackage` must be started before `AppCenterReactNativeDataPackage` so the order matters. 

We will need to figure out if there is a way to keep the order with autolinking.

jaelim-ms: Why do we have this limitation?

guperrot: We don't know when one will remove auth module. Core channel blocks on refreshing token from auth, so if SDK started without auth, channel will sign out assuming auth module was removed.
React-native splits start modules also.

PR #650, opened 2019-07-26
Added TypeScript declaration for Data

jackycodes:

## Description

[AB#65392](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/65392)

Added TypeScript declaration file for Data

### DIFF
--- a/appcenter-data/Data.d.ts
+++ b/appcenter-data/Data.d.ts
@@ -23,13 +23,13 @@ export class WriteOptions {
 }
 
 export interface DocumentWrapper {
-    deserializedValue: object;
-    eTag: string;
-    id: string;
-    isFromDeviceCache: boolean;
-    jsonValue: string;
-    lastUpdatedDate: Date;
-    partition: string;
+    deserializedValue: object | null;
+    eTag: string | null;
+    id: string | null;
+    isFromDeviceCache: boolean | null;
+    jsonValue: string | null;
+    lastUpdatedDate: Date | null;
+    partition: string | null;
 }
 
 export interface PaginatedDocuments {
--- a/appcenter-data/Data.d.ts
+++ b/appcenter-data/Data.d.ts
@@ -1,0 +1,49 @@
+export const enum TimeToLive {
+    INFINITE = -1,
+    NO_CACHE = 0,
+    DEFAULT = -1
+}
+
+export const enum DefaultPartitions {
+    USER_DOCUMENTS = 'user',
+    APP_DOCUMENTS = 'readonly'
+}
+
+export class ReadOptions {
+    timeToLive: number | TimeToLive;
+    constructor(timeToLive: number | TimeToLive);
+}
+
+export class WriteOptions {
+    timeToLive: number | TimeToLive;
+    constructor(timeToLive: number | TimeToLive);
+}
+
+export interface DocumentWrapper {
+    deserializedValue: object;
+    eTag: string;
+    id: string;
+    isFromDeviceCache: boolean;
+    jsonValue: string;
+    lastUpdatedDate: Date;
+    partition: string;
+}
+
+export interface PaginatedDocuments {
+    currentPage: Page;
+    hasNextPage: () => Promise<boolean>;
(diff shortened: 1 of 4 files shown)


appcenter-data/Data.d.ts
```diff
@@ -0,0 +1,52 @@
...
+    timeToLive: number | TimeToLive;
+    constructor(timeToLive: number | TimeToLive);
+}
+
+export interface DocumentWrapper {
+    deserializedValue: object;
+    eTag: string;
+    id: string;
+    isFromDeviceCache: boolean;
+    jsonValue: string;
```
guperrot: In strict mode, values are not nullable unless specified otherwise. We should add `| null` on some of the properties that can have a null value.

jackycodes: Added

appcenter-data/Data.d.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export class WriteOptions {
+    timeToLive: number | TimeToLive;
+    constructor(timeToLive: number | TimeToLive);
+}
+
+export interface DocumentWrapper {
+    deserializedValue: object | null;
+    eTag: string | null;
+    id: string | null;
```
guperrot: can this be ever null?

appcenter-data/Data.d.ts
```diff
@@ -0,0 +1,52 @@
...
+    isFromDeviceCache: boolean | null;
```
guperrot: can this one be ever null?

appcenter-data/Data.d.ts
jackycodes: Looks like it’s there isn’t a case where it’s null. `| null` is removed.

appcenter-data/Data.d.ts
jackycodes: Looks like it’s there isn’t a case where it’s null. `| null` is removed.

appcenter-data/Data.js
```diff
@@ -151,10 +151,15 @@ Data.WriteOptions = class {
 };
 
 function convertTimestampToDate(result) {
-    if (result && result.lastUpdatedDate) {
```
guperrot: would rather just use else { return null; }

jackycodes: Fixed

appcenter-data/Data.d.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export interface PaginatedDocuments {
+    currentPage: Page;
+    hasNextPage: () => Promise<boolean>;
+    getNextPage: () => Promise<Page>;
+    close: () => void;
+}
+
+export interface Page {
+    items: DocumentWrapper[];
```
guperrot: check if it can be null

appcenter-data/Data.js
```diff
@@ -150,12 +150,14 @@ Data.WriteOptions = class {
...
 
-function convertTimestampToDate(result) {
-    if (result && result.lastUpdatedDate) {
+function convertTimestampToDate(documentWrapper) {
+    if (documentWrapper.lastUpdatedDate) {
         // Create a new `Date` object from timestamp as milliseconds.
-        result.lastUpdatedDate = new Date(result.lastUpdatedDate);
+        documentWrapper.lastUpdatedDate = new Date(documentWrapper.lastUpdatedDate);
+    } else {
+        documentWrapper.lastUpdatedDate = null;
```
guperrot:
```suggestion
        
        // If the last updated date is 0, don't convert to January 1st 1970 UTC. Delete operation returns a timestamp of 0.
        documentWrapper.lastUpdatedDate = null;
```

jackycodes: Added

appcenter-data/Data.d.ts
jackycodes: `items` can’t be null. I changed the behaviour of `getNextPage()` to reject if there are no additional pages.

PR #658, opened 2019-08-02
Remove binaries from DemoApp

michaelsmitha:

## Description

This removes the ios/Pods folder and relies on the build script running `pod install` to download them instead.

## Related PRs or issues

[AB#65479](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/65479)

## Misc


DemoApp/appcenter-post-clone.sh
```diff
@@ -3,3 +3,6 @@ echo "Executing post clone script in `pwd`"
 echo $NPM_RC | base64 --decode > $APPCENTER_SOURCE_DIRECTORY/DemoApp/.npmrc
 # Delete everything except DemoApp folder
 rm -rf ../appcenter* ../AppCenterReactNativeShared ../TestApp34 ../BrownfieldTestApp ../TestApp
+
+pod repo add $REPO_NAME https://$USER_ACCOUNT:$ACCESS_TOKEN@msmobilecenter.visualstudio.com/SDK/_git/$REPO_NAME
```
guperrot: not sure why repo name needs to be a variable since msmobilecenter.visualstudio.com is harcoded.
Either hardcode both or variables for both.

jaelim-ms: It could be anything but we are reusing the repo name as its name on VM since it shouldn't conflict.

PR #668, opened 2019-08-13
Don't reset the auth token when sStartAutomatically flag is set

MatkovIvan:

## Description

Fix the issue when the user will be silently signed out if they restart the app

[AB#66668](https://dev.azure.com/msmobilecenter/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/66668)

### DIFF
--- a/AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
+++ b/AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
@@ -61,14 +61,16 @@ public class AppCenterReactNativeShared {
         }
 
         /*
-         * In case when sStartAutomatically flag is set, we are sure that Auth service will be started
-         * with other services, but by separately `start` call. So, we don't need to reset the auth token
-         * after first start call.
+         * When sStartAutomatically flag is set to true, every service (analytics/auth/crashes/etc.)
+         * will be started by separate AppCenter.start call. If Auth module is used,
+         * call doNotResetAuthAfterStart to avoid resetting the auth token.
          */
         try {
             Class.forName("com.microsoft.appcenter.auth.Auth");
             AuthTokenContext.getInstance().doNotResetAuthAfterStart();
         } catch (Exception ignored) {
+
+            /* Nothing to handle; this is reached if Auth SDK isn't used. */
         }
     }
 
--- a/AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
+++ b/AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
@@ -9,6 +9,7 @@ import android.text.TextUtils;
 import com.microsoft.appcenter.AppCenter;
 import com.microsoft.appcenter.ingestion.models.WrapperSdk;
 import com.microsoft.appcenter.utils.AppCenterLog;
+import com.microsoft.appcenter.utils.context.AuthTokenContext;
 
 import org.json.JSONObject;
 
@@ -57,6 +58,17 @@ public class AppCenterReactNativeShared {
         } else {
             AppCenterLog.debug(LOG_TAG, "Configure with secret.");
             AppCenter.configure(application, sAppSecret);
+        }
+
+        /*
+         * In case when sStartAutomatically flag is set, we are sure that Auth service will be started
+         * with other services, but by separately `start` call. So, we don't need to reset the auth token
+         * after first start call.
+         */
+        try {
+            Class.forName("com.microsoft.appcenter.auth.Auth");
+            AuthTokenContext.getInstance().doNotResetAuthAfterStart();
+        } catch (Exception ignored) {
         }
     }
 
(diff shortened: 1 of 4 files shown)


AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
```diff
@@ -58,6 +59,17 @@ public static synchronized void configureAppCenter(Application application) {
...
+        /*
+         * In case when sStartAutomatically flag is set, we are sure that Auth service will be started
+         * with other services, but by separately `start` call. So, we don't need to reset the auth token
+         * after first start call.
+         */
+        try {
+            Class.forName("com.microsoft.appcenter.auth.Auth");
+            AuthTokenContext.getInstance().doNotResetAuthAfterStart();
+        } catch (Exception ignored) {
+        }
```
dhei: Add a comment (example [here](https://github.com/microsoft/appcenter-sdk-android/blob/develop/apps/sasquatch/src/main/java/com/microsoft/appcenter/sasquatch/activities/SettingsActivity.java#L375)).

`/* Nothing to handle; this is reached if Auth SDK isn't used. */`

AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
```diff
@@ -58,6 +59,17 @@ public static synchronized void configureAppCenter(Application application) {
             AppCenterLog.debug(LOG_TAG, "Configure with secret.");
             AppCenter.configure(application, sAppSecret);
         }
+
+        /*
+         * In case when sStartAutomatically flag is set, we are sure that Auth service will be started
```
dhei: I suggest the following comment.

> When sStartAutomatically flag is set to true, every service (analytics/auth/crashes/etc.) will be started by separate `AppCenter.start` call. If Auth module is used, call `doNotResetAuthAfterStart` to avoid resetting the auth token.

AppCenterReactNativeShared/ios/AppCenterReactNativeShared/AppCenterReactNativeShared.m
```diff
@@ -53,6 +61,15 @@ + (void)configureAppCenter {
       } else {
         [MSAppCenter configureWithAppSecret:appSecret];
       }
+
+      /*
+       * When sStartAutomatically flag is set to true, every service (analytics/auth/crashes/etc.)
+       * will be started by separate AppCenter.start call. If Auth module is used,
+       * call doNotResetAuthAfterStart to avoid resetting the auth token.
```
jaelim-ms: This isn't `doNotResetAuthAfterStart`.

AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
```diff
@@ -58,6 +59,19 @@ public static synchronized void configureAppCenter(Application application) {
...
             AppCenter.configure(application, sAppSecret);
         }
+
+        /*
+         * When sStartAutomatically flag is set to true, every service (analytics/auth/crashes/etc.)
+         * will be started by separate AppCenter.start call. If Auth module is used,
+         * call doNotResetAuthAfterStart to avoid resetting the auth token.
+         */
+        try {
+            Class.forName("com.microsoft.appcenter.auth.Auth");
```
jaelim-ms: What if Auth class is loaded but not registered?

MatkovIvan: What case exactly do you mean?

AppCenterReactNativeShared/ios/AppCenterReactNativeShared/AppCenterReactNativeShared.m
MatkovIvan: ugrh! yeah, it's a wrong one

AppCenterReactNativeShared/android/src/main/java/com/microsoft/appcenter/reactnative/shared/AppCenterReactNativeShared.java
MatkovIvan: It isn't the case here

jaelim-ms: Talked offline. This is the only case when it auto-starts so it should be fine.

AppCenterReactNativeShared/ios/AppCenterReactNativeShared/AppCenterReactNativeShared.m
```diff
@@ -53,6 +61,15 @@ + (void)configureAppCenter {
       } else {
         [MSAppCenter configureWithAppSecret:appSecret];
       }
+
+      /*
+       * When sStartAutomatically flag is set to true, every service (analytics/auth/crashes/etc.)
```
dhei:
```suggestion
       * When startAutomatically flag is set to true, every service (analytics/auth/crashes/etc.)
```

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@
 
 * **[Feature]** Catch low memory warning and provide the API to check if it has happened in last session:  `Crashes.hasReceivedMemoryWarningInLastSession()`.
 
+### App Center Auth
+
+* **[Fix]** Fix silently signed out on application restart if auth module isn't initialized before other modules.
```
dhei: I believe the problem only happens when auth module is initialized **later** than other modules.
```suggestion
* **[Fix]** Fix silently signed out on application restart if appcenter-auth module is initialized after other appcenter modules.
```

PR #674, opened 2019-08-19
Remove outdated copy of native binaries

russelarms:

## Description

https://github.com/microsoft/appcenter-sdk-react-native/tree/develop/AppCenterReactNativeShared/ios

*.frameworks are used here only for core headers. Also, it looks very outdated.
We need to drop the copy of it and use the latest version from Pods like in other places

## Related PRs or issues

[AB#66766](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/66766)


russelarms:
I have added an additional pipeline step to install Pods (it's my testing pipeline):
https://dev.azure.com/msmobilecenter/SDK/_apps/hub/ms.vss-ciworkflow.build-ci-hub?_a=edit-build-definition&id=3531
It should be added to the main pipeline after the approval.

AppCenterReactNativeShared/ios/post-clone.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+echo "Installing pods for AppCenterReactNativeShared in `pwd`"
+
```
dhei: Please add `pod repo update` before `pod install`.

Azure Pipeline mac VM doesn't always come with the latest pod repos.

AppCenterReactNativeShared/ios/Podfile
```diff
@@ -0,0 +1,11 @@
+# Uncomment the next line to define a global platform for your project
+# platform :ios, '9.0'
+
+target 'AppCenterReactNativeShared' do
+  # Comment the next line if you don't want to use dynamic frameworks
+  use_frameworks!
```
dhei: Are we sure we want to use dynamic frameworks here? 

Does `use_frameworks!` matter here? If you comment out `use_frameworks!`, would it still work?

I don't see `use_framework!` in TestApp and DemoApp, so I am not sure if we should or shouldn't use it.

AppCenterReactNativeShared/ios/Podfile
```diff
@@ -0,0 +1,11 @@
...
+
+  # Pods for AppCenterReactNativeShared
+  pod 'AppCenter'
```
dhei: We use `pod 'AppCenter', '~> 2.3.0'` with a version number in DemoApp to allow testing pre-release private native ios sdk. It's more flexible that way. Maybe we should do the same here.

AppCenterReactNativeShared/ios/Podfile
```diff
@@ -0,0 +1,11 @@
+# Uncomment the next line to define a global platform for your project
```
dhei: Just nit-picking, I think we can get rid of these two lines of comments as we won't need it for global scope anyway.

AppCenterReactNativeShared/ios/Podfile
russelarms: It doesn't build without this line, here are the errors I got:

```
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: can't locate file for: -lPods-AppCenterReactNativeShared

error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: file: -lPods-AppCenterReactNativeShared is not an object file (not allowed in a library)
```

PR #691, opened 2019-08-29
Remove Data.setTokenExchangeUrl API from JavaScript and add the nullable check in Data.d.ts

YiMysty:

## Description

A few sentences describing the overall goals of the pull request.

## Related PRs or issues

Change the null declaration to typescript style `?`

## Misc


appcenter-data/Data.d.ts
```diff
@@ -23,13 +23,13 @@ export class WriteOptions {
 }
 
 export interface DocumentWrapper {
-    deserializedValue: object | null;
-    eTag: string | null;
+    deserializedValue?: object;
```
guperrot: that's not the same: property being undefined or defined to be null.
did you check? I think we put null explicitly, please double check.

dhei: Make sure you tested your assumption. 😄

YiMysty: changed :)

appcenter-data/Data.d.ts
```diff
@@ -48,17 +48,17 @@ export interface RemoteOperationListener {
...
 export interface RemoteOperationCompletedData {
-    id: string;
-    eTag: string;
-    partition: string;
+    id?: string;
+    eTag?: string;
+    partition?: string;
     operation: string;
-    error: Map<string, string> | null;
+    error?: Map<string, string>;
```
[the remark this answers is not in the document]
guperrot: Well we only have 1 key for now which is message so technically it works as a map but this is too generic. The type is an Error that has a message.

Would be more like error?: DataError; where DataError has a single message property of type string.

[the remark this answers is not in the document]
YiMysty: sure :D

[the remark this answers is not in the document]
dhei: Discussed with @YiMysty, `DataSyncError` sounds more appropriate here. It's more generic and easier to follow.

appcenter-data/Data.d.ts
```diff
@@ -47,18 +47,22 @@ export interface RemoteOperationListener {
...
-    error: Map<string, string> | null;
+    error?: DataSyncError;
 }
- 
+
 export function isEnabled(): Promise<boolean>;
 export function setEnabled(enabled: boolean): Promise<void>;
 export function setRemoteOperationListener(listener: RemoteOperationListener): Promise<void>;
-export function setTokenExchangeUrl(tokenExchangeUrl: string): void; 
+export function setTokenExchangeUrl(tokenExchangeUrl: string): void;
```
guperrot: We don't expose that since it cannot work on Javascript (too late).
Needs to be called in native code.

guperrot: We need to remove this from JS and bridge too, I think I told Jacky about it but this may have been lost.

dhei: Remove this API from d.ts

YiMysty: changed :)

PR #712, opened 2019-09-21
Android setAuthTokenListener API

dhei:

## Description

Android setAuthTokenListener API

Tested on RN TestApp:
- when JavaScript loaded before acquireAuthToken event
- when JavaScript loaded after acquireAuthToken event

## Related PRs or issues

[AB#67955](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/67955)


appcenter/AppCenter.js
```diff
@@ -102,6 +104,17 @@ const AppCenter = {
 
     getSdkVersion() {
         return PackageJson.version;
+    },
+
+    setAuthTokenListener(listener) {
+        eventEmitter.removeAllListeners(onAcquireAuthTokenEvent);
+        if (listener && listener.acquireAuthToken) {
```
guperrot: This contradicts typescript definition line 43. The callback is optional here.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,69 @@
...
+
+public class AppCenterReactNativeListener implements AuthTokenListener {
+
+    private static final String ON_ACQUIRE_AUTH_TOKEN_EVENT = "OnAcquireAuthTokenEvent";
+
+    private ReactApplicationContext mReactApplicationContext;
+
+    private AuthTokenCallback mCallback;
+
+    private boolean authTokenEventSent = false;
```
guperrot: mAuthTokenEventSent

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,69 @@
...
+
+    private void sendEvent(String eventType) {
+        if (mReactApplicationContext != null) {
+            if (mReactApplicationContext.hasActiveCatalystInstance() && !authTokenEventSent) {
+                mReactApplicationContext
+                        .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
+                        .emit(eventType, null);
+                authTokenEventSent = true;
+            } else {
+                replayPendingEvents();
```
guperrot: why calling that? we are not ready here, replay will call sendEvent again, this is recursive

guperrot: Here you need to save mHasPendingEvent true instead. The current boolean logic does not seem to work

appcenter/AppCenter.d.ts
```diff
@@ -37,3 +37,9 @@ export function setEnabled(enabled: boolean): Promise<void>;
 export function setUserId(userId: string | null): Promise<void>;
 export function getInstallId(): Promise<string>;
 export function setCustomProperties(properties: CustomProperties);
+export function setAuthToken(authToken: string | null): void;
+export function setAuthTokenListener(listener: AuthTokenListener) : void;
+
+export interface AuthTokenListener {
+    acquireAuthToken: () => string | null;
```
guperrot:
```suggestion
    acquireAuthToken?: () => Promise<string | null> | string | null;
```

Not sure about syntax but we are awaiting this so either a promise or the direct value works.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,69 @@
...
+    public void acquireAuthToken(AuthTokenCallback callback) {
+        ReactNativeUtils.logInfo("Auth token acquired.");
+        sendEvent(ON_ACQUIRE_AUTH_TOKEN_EVENT);
+        mCallback = callback;
+    }
+
+    private void sendEvent(String eventType) {
+        if (mReactApplicationContext != null) {
+            if (mReactApplicationContext.hasActiveCatalystInstance() && !authTokenEventSent) {
+                mReactApplicationContext
```
guperrot: We could still emit too early with that check, the real check is to wait for setAuthTokenListener to be called. So that would be another boolean instead.

appcenter/AppCenter.d.ts
guperrot: Awaiting on primitive types work, so technically we actually support both promise and directly the value.

The comment is thus not addressed (also the optional aspect of it is still true).

appcenter/AppCenter.js
guperrot: As discussed, if we are keeping required in TS, we should emit a AppCenterLog.error if listener is not null but callback missing.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,62 @@
...
+
+public class AppCenterReactNativeListener implements AuthTokenListener {
+
+    private static final String ON_ACQUIRE_AUTH_TOKEN_EVENT = "OnAcquireAuthTokenEvent";
+
+    private ReactApplicationContext mReactApplicationContext;
+
+    private AuthTokenCallback mCallback;
+
+    private boolean mAuthTokenEventSent = false;
```
guperrot: Setting fields explicitly to null, false or 0 is usually not what we do in our Java/C# code.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,62 @@
...
+    private boolean mHasPendingEvent = false;
+
+    @SuppressWarnings("WeakerAccess")
+    public final void setReactApplicationContext(ReactApplicationContext reactApplicationContext) {
+        mReactApplicationContext = reactApplicationContext;
+    }
+
+    @Override
+    public void acquireAuthToken(AuthTokenCallback callback) {
+        ReactNativeUtils.logInfo("Auth token requested from native code.");
```
guperrot: debug

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,62 @@
...
+        sendEvent(ON_ACQUIRE_AUTH_TOKEN_EVENT);
+        mCallback = callback;
+    }
+
+    private void sendEvent(String eventType) {
+        if (mReactApplicationContext != null) {
```
guperrot: Is that a dead condition? If it's real we need to switch boolean too for later. If that condition is never null then we can remove it from code.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,62 @@
...
+            if (mHasPendingEvent) {
```
guperrot: If JS is ready already, and we have to refresh token, this value is false, thus it does not work.
We need to set mHasPendingEvent to true when acquireAuthToken is called by native SDK. Then call `sendEvent`.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,62 @@
...
+                        .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
+                        .emit(eventType, null);
+                mAuthTokenEventSent = true;
+            } else {
+                mHasPendingEvent = true;
+            }
+        }
+    }
+
+    // Replay method is only called when JavaScript SetAuthTokenListener is registered.
```
guperrot: slash star comments

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,62 @@
...
+        mCallback = callback;
+    }
+
+    private void sendEvent(String eventType) {
+        if (mReactApplicationContext != null) {
+            if (mHasPendingEvent) {
+                mReactApplicationContext
+                        .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
+                        .emit(eventType, null);
+                mAuthTokenEventSent = true;
```
guperrot: After emitting, just switch mHasPendingEvent to false and then you don't need the second boolean.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
guperrot: This check should actually be:
```suggestion
        if (mAuthTokenEventListenerSet) {
```
Which would mean removing `mAuthTokenEventSent` and using `mAuthTokenEventListenerSet` instead.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,62 @@
...
+            } else {
+                mHasPendingEvent = true;
+            }
+        }
+    }
+
+    // Replay method is only called when JavaScript SetAuthTokenListener is registered.
+    void replayPendingEvents() {
+        if (!mAuthTokenEventSent) {
```
guperrot: Instead of this if, consider `mAuthTokenEventListenerSet = true`, then call `sendEvent`.
Might need to rename the method also.

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @SuppressWarnings("WeakerAccess")
+    public final void setReactApplicationContext(ReactApplicationContext reactApplicationContext) {
+        mReactApplicationContext = reactApplicationContext;
+    }
+
+    @Override
+    public void acquireAuthToken(AuthTokenCallback callback) {
+        ReactNativeUtils.logInfo("Auth token requested from native code.");
+        sendEvent(ON_ACQUIRE_AUTH_TOKEN_EVENT);
```
guperrot:
```suggestion
        mHasPendingEvent = true;
        sendEvent(ON_ACQUIRE_AUTH_TOKEN_EVENT);
```

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
```diff
@@ -27,29 +27,24 @@ public final void setReactApplicationContext(ReactApplicationContext reactApplic
...
-                        .emit(eventType, null);
-                mAuthTokenEventSent = true;
-            } else {
-                mHasPendingEvent = true;
-            }
+    private void sendEvent(String eventType, Object data) {
+        if (mAuthTokenEventListenerSet && mHasPendingEvent) {
+            mReactApplicationContext
+                    .getJSModule(DeviceEventManagerModule.RCTDeviceEventEmitter.class)
+                    .emit(eventType, data);
```
guperrot: add mHasPending = false after emitting.
```suggestion
                    .emit(eventType, data);
        mHasPendingEvent = false;
```

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
dhei: Keeping the `replayPendingEvents` method name because it still makes sense.

appcenter/AppCenter.js
guperrot: Bump

appcenter/android/src/main/java/com/microsoft/appcenter/reactnative/appcenter/AppCenterReactNativeListener.java
dhei: `if (mReactApplicationContext != null)` is a dead condition, so we removed the check.

appcenter/AppCenter.js
```diff
@@ -102,6 +104,24 @@ const AppCenter = {
 
     getSdkVersion() {
         return PackageJson.version;
+    },
+
+    setAuthTokenListener(listener) {
+        eventEmitter.removeAllListeners(onAcquireAuthTokenEvent);
+        if (listener) {
+            if (listener.acquireAuthToken) {
```
guperrot:
```suggestion
            if (typeof(listener.acquireAuthToken) === "function") {
```

PR #718, opened 2019-09-26
Fix changelog for 2.4.0 release

dhei:

## Description
Because this is the first time we ship App Center Data, an API change on App Center Apple SDK shouldn't be identified as "breaking change" at all. Remove it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,11 +25,9 @@
 
 ### App Center Data
 
- The App Center Data service provides functionality enabling developers to persist app data in the cloud in both online and offline scenarios. This enables you to store and manage both user-specific data as well as data shared between users and across platforms.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -25,11 +25,9 @@
 
 ### App Center Data
 
- The App Center Data service provides functionality enabling developers to persist app data in the cloud in both online and offline scenarios. This enables you to store and manage both user-specific data as well as data shared between users and across platforms.
+Version 2.4.0 of the App Center React Native SDK includes a new module: Data.
```
jaelim-ms: Doesn't seem consistent with previous messages.

jaelim-ms: Do we need this line? You don't need to mention `Version 2.4.0` in 2.4.0 change log. 🤔

PR #740, opened 2019-10-07
Fix creating document bug

tjyung:

## Description

Creating a new document doesn't work because the ttl input was a string when it needed to be an int.

## Related PRs or issues
[#AB69661](https://msmobilecenter.visualstudio.com/Mobile-Center/_boards/board/t/Yoann-Team/Backlog%20Items/?workitem=69661)

### DIFF
--- a/appcenter-data/Data.d.ts
+++ b/appcenter-data/Data.d.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 export const enum TimeToLive {
+    INFINITE = -1,
     NO_CACHE = 0,
     DEFAULT = -1
 }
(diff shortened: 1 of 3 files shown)


appcenter-data/Data.d.ts
```diff
@@ -2,7 +2,6 @@
 // Licensed under the MIT License.
 
 export const enum TimeToLive {
-    INFINITE = -1,
```
jwallra: Why is this removed?

tjyung: The default is now infinite so was doing some cleanup but I'll put it back because its still used somewhere.

PR #741, opened 2019-10-07
Change exception to use stack trace in android

achocron:

## Description

[AB#69645](https://msmobilecenter.visualstudio.com/454a20a9-dfe6-4c1f-8ae8-417f76adb472/_workitems/edit/69645)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@
   - Accounts in any Azure AD directory
   - Accounts in any Azure AD directory and personal Microsoft accounts
 
+### App Center Crashes
+
+#### Android
+* **[Behavior Change]** Fixes a security issue in the way native Android crashes are processed. The exception message will now be `null` and the exception stack trace might look slightly different.
+
 ___
 
 ## Version 2.4.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@
 ### App Center Crashes
 
 #### Android
-* **[Behavior Change]** Fixes a security issue in the way native Android crashes are processed. The exception message will now be `null` and the exception stack trace might look slightly different.
+
+* **[Behavior change]** Fix a security issue in the way native Android crashes are processed. The exception message is now `null` and the exception stack trace is the raw Java stack trace.
 
 ___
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
   - Accounts in any Azure AD directory
   - Accounts in any Azure AD directory and personal Microsoft accounts
 
+### App Center Crashes
+
+#### Android
+* **[Behavior Change]** Fixes a security issue in the way native Android crashes are processed. The exception message will now be `null` and the exception stack trace might look slightly different.
```
guperrot:
```suggestion

* **[Behavior Change]** Fix a security issue in the way native Android crashes are processed. The exception message is now `null` and the exception stack trace is the raw Java stack trace.
```

appcenter-crashes/android/src/main/java/com/microsoft/appcenter/reactnative/crashes/AppCenterReactNativeCrashesUtils.java
```diff
@@ -41,13 +41,11 @@ static WritableMap convertErrorReportToWritableMap(ErrorReport errorReport) thro
...
         errorReportMap.putString("appStartTime", "" + errorReport.getAppStartTime().getTime());
-        
-        //noinspection ThrowableResultOfMethodCallIgnored
-        Throwable error = errorReport.getThrowable();
-        if (error != null) {
-            errorReportMap.putString("exception", Log.getStackTraceString(error));
-            errorReportMap.putString("exceptionReason", error.getMessage());
+        String stackTrace = errorReport.getStackTrace();
+        if (stackTrace != null) {
+            errorReportMap.putString("exception", stackTrace);
```
jwallra: Wouldn't this change cause a test to fail?  Do any tests need to be added?

michaelsmitha: I think either way you are just putting a string in to "exception" of the errorReportMap. Not sure if there is already a test for this, but seems likely it would continue to pass with this change

jwallra: Well, we know the code is being covered, but my suspicion is that the contents of the entry in the map are not checked

dhei: @jwallra @michaelsmitha We have no unit tests for RN Java or Objective-C bridge code (only tests for JavaScript). Filling the test gap is something we want to do but never got to it.

PR #751, opened 2019-10-22
Add exact version of rnfs

annakocheshkova:
In latest versions of RN, `RCTImageLoader.h` import was deprecated. RNFS module created a patch replacing this import with a new one, `RCTImageLoaderProtocol.h`. But we test backwards compatibility on RN 0.59.9 and it breaks archive for us.

Therefore, I had to specify the exact `RNFS` version *without* this change.


annakocheshkova:
Merging this because can't verify from a branch not starting with "release"

PR #762, opened 2019-11-01
Fix TestApp DataScreen issues

russelarms:

## Description

Fixed:
- After creating a user document, the newly created document isn't shown in the user documents list, unless we select Partition to "User documents" again.
- The user documents are still shown in the list after user sign out
- The previous inputted values still exist in creating document page when click "Created a new document"
- Confusing "Create Document" layout

## Related PRs or issues

[AB#70478](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/70478)
[AB#70476](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/70476)
[AB#70475](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/70475)

## Misc

### DIFF
--- a/DemoApp/app/screens/DataScreen.js
+++ b/DemoApp/app/screens/DataScreen.js
@@ -68,9 +68,9 @@ export default class DataScreen extends Component {
       docType: '',
       docKey: '',
       docValue: '',
-      documents: documents,
+      documents,
       loadingData: false
-    });  
+    });
   }
 
   async listDocuments(partition) {
@@ -284,10 +284,10 @@ export default class DataScreen extends Component {
                               new Data.WriteOptions(this.state.docTtl)
                             );
                     console.log('Successful create', createResult);
-                    this.setState({ createDocModalVisible: !this.state.createDocModalVisible }); 
+                    this.setState({ createDocModalVisible: !this.state.createDocModalVisible });
                     if (!this.state.createDocModalVisible) {
                       await this.hideCreateDocModal();
-                    }       
+                    }
                   }}
              >
                <Text style={[SharedStyles.itemButton]}>Create</Text>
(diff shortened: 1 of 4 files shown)


DemoApp/app/screens/DataScreen.js
```diff
@@ -54,10 +54,26 @@ export default class DataScreen extends Component {
...
     const dataEnabled = await Data.isEnabled();
     this.setState({ dataEnabled });
+    if (!this.state.loadingData) {
+      const documents = await this.listDocuments(this.state.partition);
+      this.setState({ documents, loadingData: false });
+    }
   }
 
-  setCreateDocModalVisible(visible) {
+  async setCreateDocModalVisible(visible) {
```
dhei: The `setCreateDocModalVisible` method and how the method is called is very confusing. I suggest we refactor the code to make it easier to follow.
 
In the current PR, when we successfully created the new document, `setCreateDocModalVisible` is called with  `!createDocModalVisible` parameter (https://github.com/microsoft/appcenter-sdk-react-native/pull/762/files#diff-5c743115ed0a0fcdd754f9eac490cb37R290). 

The confusing parts are:
-  the method name `setCreateDocModalVisible` doesn't make sense if the implementation switch whatever the visibility state between true and false.
- this method only calls `setState` when `visible` parameter is false, which is the reverse value of `createDocModalVisible`.

DemoApp/app/screens/DataScreen.js
```diff
@@ -54,10 +54,26 @@ export default class DataScreen extends Component {
...
     this.setState({ createDocModalVisible: visible });
+    if (!visible) {
+      const documents = await this.listDocuments(this.state.partition);
+      this.setState({
+        docTtl: 60,
+        docId: '',
+        docType: '',
+        docKey: '',
+        docValue: '',
+        documents,
```
dhei: For readability and consistency, I think the named parameter would be better.
```suggestion
        documents: documents,
```

MatkovIvan: @dhei does `lint` agree with you?

dhei: https://eslint.org/docs/rules/object-shorthand
You are right, eslint encourages using object shorthand in this case.

DemoApp/app/screens/DataScreen.js
dhei: I like the refactored code 👍

Issue #749, opened 2019-10-21
Cannot compile project with TypeScript `--isolatedModules` enabled


newyankeecodeshop:
### **Description**

I am trying to build our React Native project using the guidelines described here:
https://facebook.github.io/react-native/docs/typescript#adding-typescript-to-an-existing-project
The major issue is when enabling `isolatedModules`. Our code imports the following from the AppCenter React Native SDK:
```
import AppCenter from "appcenter";
import Crashes, { ErrorAttachmentLog } from "appcenter-crashes";
```

When I run the TypeScript compiler, I get the following errors:

```
yarn run v1.15.2
$ tsc
node_modules/appcenter-crashes/Crashes.d.ts:1:19 - error TS1209: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.

1 export const enum UserConfirmation {
                    ~~~~~~~~~~~~~~~~

node_modules/appcenter/AppCenter.d.ts:1:19 - error TS1209: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.

1 export const enum LogLevel {
                    ~~~~~~~~

node_modules/appcenter/AppCenter.d.ts:14:19 - error TS1209: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.

14 export const enum AppCenterLogLevel {
                     ~~~~~~~~~~~~~~~~~

error Command failed with exit code 1.
```

From my searching online, it appears that library declarations need to avoid `const enum` so they can be used with or without the `isolatedModules` flag. Since this flag is now recommended by the React Native TypeScript guidelines, it is something the AppCenter SDK needs to support.

### **Repro Steps**

Please list the steps used to reproduce your issue.

1. Create a React Native project (0.57 or later) and enable the TypeScript integration
2. Add a .ts file that imports artifacts from "appcenter" and "appcenter-crashes".
3. Typecheck the project (e.g. `yarn tsc`)

### **Details**

1. Which version of the App Center SDK are you using?
    1.12.0
2. Which OS version did you experience the issue on?
    N/A
3. What device version did you see this error on?  Were you using an emulator or a physical device?
    N/A
4. What third party libraries are you using?
    React Native 0.57, TypeScript 3.1
5. Run the following command and paste the output below: `react-native info`

```
  React Native Environment Info:
    System:
      OS: macOS 10.14.6
      CPU: x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
[...]

PR #767, opened 2019-11-11
Fix/tscompilererrors

nisteuve:

## Description

There were typescript compiler errors due to two factors:

-Use of "const enum" rather than "enum"
-Missing return type for several exported functions

## Related PRs or issues

https://github.com/microsoft/appcenter-sdk-react-native/issues/749


CHANGELOG.md
```diff
@@ -2,10 +2,21 @@
 
 ## Version 2.5.1 (Under development)
 
+### App Center
+
+* **[Fix]** Fix typescript compiler errors related to const enum usage and missing return types
```
guperrot:
```suggestion
* **[Fix]** Fix typescript compiler errors related to const enum usage and missing return types.
```

CHANGELOG.md
```diff
@@ -2,10 +2,21 @@
...
+
 ### App Center Analytics
 
 * **[Fix]** Avoid throwing an error when `trackEvent` is called with null property values.
 
+### App Center Crashes
+
+* **[Fix]** Fix typescript compiler errors related to const enum usage
```
guperrot:
```suggestion
* **[Fix]** Fix typescript compiler errors related to const enum usage.
```

CHANGELOG.md
```diff
@@ -2,10 +2,21 @@
...
+
+### App Center Data
+
+* **[Fix]** Fix typescript compiler errors related to const enum usage
```
guperrot:
```suggestion
* **[Fix]** Fix typescript compiler errors related to const enum usage.
```

appcenter-data/__tests__/Data.js
```diff
@@ -77,12 +77,13 @@ describe('App Center Data read operation tests', () => {
...
 
 describe('App Center Data list operation tests', () => {
-  test('list is called ', async () => {
+  test('list is called with partition and TimeToLive.DEFAULT', async () => {
     const partition = Data.DefaultPartitions.USER_DOCUMENTS;
     const spy = jest.spyOn(NativeModules.AppCenterReactNativeData, 'list');
     const result = await Data.list(partition);
     console.log(result);
-    expect(spy).toHaveBeenCalledWith(partition);
+    const readOptions = { "timeToLive": Data.TimeToLive.DEFAULT };
```
guperrot: undefined should map to default, we are supposed to have code to handle the default case
if you had to update the unit test then it might hide a real bug in the SDK

dhei: @guperrot `await Data.list(partition)` is called without TTL parameter (line 82) and the SDK correctly treats it `TimeToLive.DEFAULT`. The problem was this test case was failing on develop branch but unit tests of Data module and Auth module are not executed in the CI. So this PR fixed a bug in the unit test, and now all the JS unit tests are executed in the CI.

appcenter-data/__tests__/Data.js
```diff
@@ -77,12 +77,13 @@ describe('App Center Data read operation tests', () => {
 });
 
 describe('App Center Data list operation tests', () => {
-  test('list is called ', async () => {
+  test('list is called with partition and TimeToLive.DEFAULT', async () => {
```
dhei: As mentioned in https://github.com/microsoft/appcenter-sdk-react-native/pull/767/files#r344940943, there is a gap in the test coverage. `Data.list` API takes partition and readOptions as parameters but we don't have a test case covering list API called with both parameters.

guperrot:
Build error is

```
/Users/runner/runners/2.160.0/work/1/s/appcenter-data/__tests__/Data.js
  85:27  error  Unnecessarily quoted property 'timeToLive' found  quote-props
  85:27  error  Strings must use singlequote                      quotes
```

appcenter-data/__tests__/Data.js
```diff
@@ -77,12 +77,13 @@ describe('App Center Data read operation tests', () => {
...
     const partition = Data.DefaultPartitions.USER_DOCUMENTS;
     const spy = jest.spyOn(NativeModules.AppCenterReactNativeData, 'list');
     const result = await Data.list(partition);
     console.log(result);
-    expect(spy).toHaveBeenCalledWith(partition);
+    const readOptions = { 'timeToLive': Data.TimeToLive.DEFAULT };
```
dhei:
```
/Users/runner/runners/2.160.0/work/1/s/appcenter-data/__tests__/Data.js
  85:27  error  Unnecessarily quoted property 'timeToLive' found  quote-props
```

MatkovIvan:
@dhei It's too late, but

I don't think that changing `const enum` to `enum` is the right solution. I placed some references in the comment to work item a while ago (when I moved it to triage), but it looks like it was missed some how.

See https://lukasbehal.com/2017-05-22-enums-in-declaration-files/
**TLDR**: Do not define your enums in your TypeScript Declaration (`*.d.ts`) files, ever.
A union type of literals should be used instead.

PR #814, opened 2020-03-19
Add explicit react-native-screens dependency to the TestApp

russelarms:

* [ ] ~~Has `CHANGELOG.md` been updated?~~

## Description

**react-navigation-tabs** package has **react-native-screens** as a [peerDependency](https://flaviocopes.com/npm-peer-dependencies/), which means it is not downloaded automatically when using `npm i`. It used to work because **react-native-screens** was required by **react-navigation/native@3.3.0**
```
"@react-navigation/native": {
"version": "3.3.0",
"resolved": "https://registry.npmjs.org/@react-navigation/native/-/native-3.3.0.tgz",
...
"requires": {
"hoist-non-react-statics": "^3.0.1",
"react-native-safe-area-view": "^0.13.0",
"react-native-screens": "^1.0.0 || ^1.0.0-alpha"
},
```

But **react-navigation/native@3.6.5** doesn't have an explicit dependency on **react-native-screens** anymore:
```
"@react-navigation/native": {
"version": "3.6.5",
"resolved": "https://registry.npmjs.org/@react-navigation/native/-/native-3.6.5.tgz",
...
"requires": {
"hoist-non-react-statics": "^3.3.2",
"react-native-safe-area-view": "^0.14.8"
},

```

## Related PRs or issues

[AB#78100](https://msmobilecenter.visualstudio.com/Mobile-Center/_workitems/edit/78100)

### DIFF
--- a/TestApp/ios/Podfile.lock
+++ b/TestApp/ios/Podfile.lock
@@ -377,7 +377,7 @@ SPEC CHECKSUMS:
   appcenter-core: 98c617ae028c90020e404f020c37d7cb189adf4c
   appcenter-crashes: add23a3c3cf0403e1f54c39d85e9b3c5db7279e7
   appcenter-push: 9a88b078ad61718947c92561b5b86da54e185582
-  AppCenterReactNativeShared: d030cbd4c11902dcff4a9cea80dac66714cc2df6
+  AppCenterReactNativeShared: c57d2566896ba19a6193046c8966691c41b4be93
   boost-for-react-native: 39c7adb57c4e60d6c5479dd8623128eb5b3f0f2c
   DoubleConversion: 5805e889d232975c086db112ece9ed034df7a0b2
   FBLazyVector: 5bc5b1606fc9a7ac6956de049f6e30901ed31c49
--- a/TestApp/ios/Podfile.lock
+++ b/TestApp/ios/Podfile.lock
@@ -249,6 +249,8 @@ PODS:
     - React
   - RNGestureHandler (1.5.0):
     - React
+  - RNScreens (2.3.0):
+    - React
   - Yoga (1.14.0)
 
 DEPENDENCIES:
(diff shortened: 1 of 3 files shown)


MatkovIvan:
It looks like revert of #806

TestApp/ios/Podfile.lock
```diff
@@ -372,7 +377,7 @@ SPEC CHECKSUMS:
   appcenter-core: 98c617ae028c90020e404f020c37d7cb189adf4c
   appcenter-crashes: add23a3c3cf0403e1f54c39d85e9b3c5db7279e7
   appcenter-push: 9a88b078ad61718947c92561b5b86da54e185582
-  AppCenterReactNativeShared: c57d2566896ba19a6193046c8966691c41b4be93
+  AppCenterReactNativeShared: d030cbd4c11902dcff4a9cea80dac66714cc2df6
```
annakocheshkova: Is this required?

russelarms:
@MatkovIvan It is needed anyway, without it, the app won't build.
The difference between the two is that DemoApp uses RN 0.59 and TestApp uses RN 0.61, they use different build flows.
It was linked the wrong way on DemoApp, that's why we applied [#806](https://github.com/microsoft/appcenter-sdk-react-native/pull/806)

TestApp/ios/Podfile.lock
dhei: @annakocheshkova @russelarms 
Checksum changes every time your build xcode project, which means, unfortunately, it will change every time we run `TestApp/update-npm-package.sh` during normal development flow even there is no change regarding iOS.
The checksum of `AppCenterReactNativeShared` is not providing much value to us, but I am fine to live with it despite of the annoyance.

TestApp/ios/Podfile.lock
```diff
@@ -288,13 +290,14 @@ DEPENDENCIES:
...
   - RNGestureHandler (from `../node_modules/react-native-gesture-handler`)
+  - RNScreens (from `../node_modules/react-native-screens`)
   - Yoga (from `../node_modules/react-native/ReactCommon/yoga`)
 
 SPEC REPOS:
-  https://cdn.cocoapods.org/:
-    - AppCenter
   https://github.com/CocoaPods/Specs.git:
     - boost-for-react-native
+  trunk:
```
guperrot: curious about that change moving away from the CDN

dhei: `trunk` is a different name for the same main spec repo backed by a CDN, please see https://github.com/CocoaPods/CocoaPods/issues/9190#issuecomment-535330760:

> trunk is the new name for the main specs repo which is backed by a CDN. This repo is included by default, and since your Podfile does not have any custom spec repos specified, it is being included in this case.

PR #833, opened 2020-05-25
Setup DemoApp for release 3.0.3

charmian01:
## Description

Setup DemoApp for testing for React Native 3.0.3 release.


DemoApp/android/build.gradle
```diff
@@ -30,6 +30,7 @@ allprojects {
         maven {
             url  "https://dl.bintray.com/vsappcenter/appcenter-snapshot"
         }
+        maven { url 'https://www.jitpack.io' }
```
MatkovIvan, line 33: I see no new deps, so I doubt that we need it

charmian01: Removed this.

PR #921, opened 2021-06-24
React Native Release 4.2.0

russelarms:

## Description

Bumped the versions
Updated TestApp to use latest SDKs
Updated the CHANGELOG.md
Fix Xcode 12.5 build for TestApp


TestApp/update-npm-packages.sh
```diff
@@ -9,18 +9,22 @@ npm pack ../appcenter
 npm pack ../appcenter-analytics
 npm pack ../appcenter-crashes
 npm pack ../appcenter-link-scripts
-# workaround for macs on arm64
-npm install react-native-fs --save --legacy-peer-deps
 
 echo "Installing appcenter* packages..."
 npm install appcenter*.tgz
 
+echo "workaround for macs on arm64"
```
DmitriyKirakosyan, line 16: do we need to "echo" it? and why did you move it?

TestApp/update-npm-packages.sh
```diff
@@ -9,18 +9,22 @@ npm pack ../appcenter
...
+npm install react-native-fs --save --legacy-peer-deps
+
 echo "Cleanup appcenter*.tgz"
 rm appcenter*.tgz
 
 echo "Installing other packages..."
 npm install
 
+echo "Apply patch to react-native package that fix build on Xcode 12.5..."
```
DmitriyKirakosyan, line 25:
```suggestion
echo "Applying patch to react-native package that fixes build on Xcode 12.5..."
```

TestApp/update-npm-packages.sh
russelarms: Echo helps to see at which step the script fails if so. I moved it because on this step we already have our package versions updated (archives with 4.2.0 packages) and `npm install` tries to install previous appcenter versions (4.1.0 in this case, they are set in the package.json at this moment) and can't find them.

TestApp/update-npm-packages.sh
```diff
@@ -9,18 +9,22 @@ npm pack ../appcenter
...
+npm install react-native-fs --save --legacy-peer-deps
+
 echo "Cleanup appcenter*.tgz"
 rm appcenter*.tgz
 
 echo "Installing other packages..."
 npm install
 
+echo "Applying patch to react-native package that fixes build on Xcode 12.5..."
+npx patch-package
```
DmitriyKirakosyan, line 26: Let's add a comment with a link to the react-native issue.
Once the fix is released in stable version we can consider upgrading react-native and remove this line.

DmitriyKirakosyan (approved):

